Issue #4, opened 2012-04-19
gui_fit includes "bad" telemetry


jeanconn:
RFE: A default or option to exclude bad data (especially bad OBC data in safe model)

taldcroft:
Closed by #27.

PR #27, opened 2015-02-24
Add bad_times model attributes to mask out resids during bad times

taldcroft:
This allows adding the following to a JSON model file to define time intervals that are ignored in fitting:
```
 {
   "bad_times": [
        [
            "2014:001", 
            "2014:003"
        ], 
        [
            "2014:010", 
            "2014:013"
        ]
    ], 
   "comps": [
        {
            "class_name": "Mask",  
      ....
 ```
Technically, the fit residuals are not ignored but are zeroed out.


taldcroft:
Closes #4.

PR #29, opened 2015-03-13
Make gui_fit recenter epoch on fit window by default for linear models

taldcroft:
This should make the typical model fitting process easier by recentering the model `epoch` for `SolarHeat`-based models that have linear long-term dependence.  This behavior is the default, but it can be disabled by passing the `--keep-epoch` flag when calling `gui_fit.py`.

A typical fitting sequence could then be:

1. Fit the solar heat `P` parameters for the most recent year of data, letting `gui_fit` recenter the epoch.  In this stage leave the `dP` values at the previous best fit values (do not zero the out).  Save and close.
2. Ft the solar heat `dP` parameters for the most recent 2-3 years, using the `--keep-epoch` flag to maintain the epoch at the center of the 1-year fit.  In this way the epoch refers to the center of the data used to determine the solar heat `P` values.

@matthewdahmer @jeanconn

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -86,6 +86,46 @@ class SolarHeat(PrecomputedHeatPower):
         self.n_mvals = 1
         self.var_func = getattr(self, var_func)
 
+    @property
+    def epoch(self):
+        return self._epoch
+
+    @epoch.setter
+    def epoch(self, value):
+        if hasattr(self, '_epoch'):
+            if self.var_func is not self.linear:
+                raise AttributeError('can only reset the epoch for var_func=linear')
+
+            new_epoch = DateTime(value)
+            epoch = DateTime(self.epoch)
+            days = new_epoch - epoch
+
+            # Don't make tiny updates to epoch
+            if abs(days) < 10:
+                return
+
+            # Update the Ps params in place.  Note that self.Ps is basically for
+            # setting the array size whereas the self.pars vals are the actual values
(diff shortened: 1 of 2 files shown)


taldcroft:
@matthewdahmer - are you good with this PR?

matthewdahmer:
My only concern is with the case where the P values are scaled outside of the range specified in the min and max solarheat parameters. In most cases this will not be a problem, however if the DP values are large enough compared to the P values, with sufficiently narrow min/max bounds on the P values, this could be a minor problem. In my testing, I found that if a P value is scaled outside the specified min/max bound, then the value is still set to the scaled value (which is good), however it then is up to the user to remain conscious of this possibility and adjust the bounds before attempting another fit. If another fit is attempted without adjusting the bounds and a parameter's scaled value is outside the specified bounds, regardless of whether or not the parameter is frozen, that parameter's value will shift to the nearest bound after pressing the fit button.

I don't see this as a fatal flaw so if you are looking to get this buttoned up so that it can be included in with the upcoming FOT tools release I am ok with merging this. I might be able to add this functionality myself, just not quite at the moment.

matthewdahmer:
So I got to thinking about this and it turns out automatically scaling the P parameter bounds is not nearly as complex as I had imagined it might be. My commit only adds a few lines of code that scales the min or max bound to match the scaled value if the scaled value is outside the originally specified bounds.

taldcroft:
@matthewdahmer - awesome.  Merging now.

PR #30, opened 2015-03-13
Make testing work properly from py.test and ipython

taldcroft:
This should also serve as a reference implementation for getting pytest test suite to work from the command line and by importing the package and doing `xija.test()`.


taldcroft:
Also, did a bunch of tidying up to put non-pytest files used within development into a single directory.

PR #39, opened 2016-01-12
Verify that spec-file par names match class-generated par names

taldcroft:
Need to make sure this does not impact any existing flight models.  Also need to check with model inheritance code.


matthewdahmer:
@taldcroft 
Would this impact model element names that use the msid as a part of the name? For example the acis focal plane model uses fptemp instead of the actual msid name fptemp_11 (this does cause an issue in pyger). I would expect an automated system to produce names using the full msid name (e.g. coupling__fptemp_11__1cbat) where hand editing could use alternative forms (e.g. coupling__fptemp__1cbat).

taldcroft:
Not worth the validation testing and potential problems.

PR #41, opened 2016-02-29
Code updates which came up while developing the tutorial update

taldcroft:
While going through the code and looking at the examples, I saw some opportunities for improvements.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -248,6 +248,16 @@ class SolarHeat(PrecomputedHeatPower):
             ax.set_xlim(40, 180)
             ax.grid()
 
+    def plot_solar_heat__time(self, fig, ax):
+        lines = ax.get_lines()
+        if not lines:
+            plot_cxctime(self.model.times, self.dvals, '-b', fig=fig, ax=ax)
+            ax.grid()
+            ax.set_title('{}: data'.format(self.name))
+            ax.margins(0.05)
+        else:
+            lines[0].set_data(self.model_plotdate, self.dvals)
+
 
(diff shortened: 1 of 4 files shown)


xija/component/heat.py
```diff
@@ -128,6 +128,8 @@ def __init__(self, model, node, pitch_comp, eclipse_comp=None,
             dPs = np.zeros_like(self.P_pitches)
         self.dPs = np.array(dPs, dtype=np.float)
 
+        if epoch is None:
```
taldcroft: This needs to be verified with all flight models since it could cause breakage.  The PSMC model does not currently have `epoch` set in the spec file, but it also does not use `SolarHeat`.

Gregg140:
I have a general question regarding adding pitch bins in a .json file:

This is a fragment of a json file I'm working with.  I want to ad a 120 degree pitch bin - which I've done in P_pitches:

...................
            "init_kwargs": {
                "P_pitches": [
                    45, 
                    60, 
                    90, 
                    105, 
		    120,
                    130, 
                    150, 
                    180
                ], 
                "Ps": [
                    0.58, 
                    0.5, 
                    0.41, 
                    0.7, 
                               <- Need to add value here
                    1.0,
		    1.0, 
                    0.9, 
                    0.79
                ], .................

But I also need to add a value in Ps. Best I can get from the documentation is that these are 'default values'. They seem to be less than one. Can you describe how they are used and how one would arrive at a default value when adding another pitch bin?

Thanks

taldcroft:
Overtaken by events.

PR #43, opened 2016-10-21
Modernize testing and fix bug in testing file path

### LINKED ISSUE #37: test() tries to write minusz.npz back into python site-packages directory

taldcroft:
Closes #37

### DIFF
--- a/xija/tests/test_models.py
+++ b/xija/tests/test_models.py
@@ -137,7 +137,7 @@ def test_minusz():
     mdl.make()
     mdl.calc()
 
-    regrfile = 'minusz.npz'
+    regrfile = abs_path('minusz.npz')
     if not os.path.exists(regrfile):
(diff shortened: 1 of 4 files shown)


xija/tests/test_models.py
```diff
@@ -137,7 +137,7 @@ def test_minusz():
     mdl.make()
     mdl.calc()
 
-    regrfile = 'minusz.npz'
+    regrfile = abs_path('minusz.npz')
```
taldcroft: Yes.  The file is bundled and is required to exist.  The existing code makes it easy to regenerate the file if needed.  In the unlikely event of some strange process failure where it gets out of the distribution then we will get the same test fail again.

jeanconn: Right, I just thought maybe the code to make it should be someplace else and/or should never try to write the file directly to arch.

PR #44, opened 2016-12-29
Python 3 compatibility

taldcroft:


jzuhone:
I looked over the source, it seems like you may want to add `from __future__ import print_function` in a few more source files.

taldcroft:
Thanks @jzuhone.  Those should all be of the form `print(one_thing)`, but I'll go ahead and do that anyway.

I hope you're psyched about Python 3.  I spent the week getting much of core Ska working on Py3 except using Py 3.5 because 3.6 isn't fully available in conda).

Next on my list is trying out the ACIS thermal checkers...

jeanconn:
:+1: 

Only tested from the 2.18 environment.

jzuhone:
@taldcroft do you think this PR needs anything else? or can we merge it?

taldcroft:
Yikes, I had completely forgotten about this.  What's worse is that 3.7.1 is actually installed to flight Ska.  So this should indeed be merged, but I just need to double-check that this PR exactly corresponds to what is installed.  Standby.

taldcroft:
OK, confirmed that the 3.7.1 version in /proj/sot/ska/pkgs/ (and Ska flight) is exactly this PR.

PR #50, opened 2018-04-04
Support the (soon-to-be) latest version of taco

jzuhone:
In the next version of `taco`, it will be a top-level package. This PR makes sure that regardless of whether `taco` is installed or `Chandra.taco` is installed, it can be used with `xija`.

### DIFF
--- a/xija/gui_fit.py
+++ b/xija/gui_fit.py
@@ -28,7 +28,10 @@ import sherpa.ui as ui
 from Chandra.Time import DateTime
 import pyyaks.context as pyc
 
-import Chandra.taco
+try:
+    import taco
+except ImportError:
+    import Chandra.taco as taco
 import xija
 import xija.clogging as clogging   # get rid of this or something
 
@@ -662,7 +665,7 @@ def get_options():
 def main():
     # Enable fully-randomized evaluation of ACIS-FP model which is desirable
     # for fitting.
-    Chandra.taco.taco.set_random_salt(None)
+    taco.taco.set_random_salt(None)
 
     opt = get_options()
 
--- a/xija/gui_fit.py
+++ b/xija/gui_fit.py
@@ -29,7 +29,7 @@ from Chandra.Time import DateTime
 import pyyaks.context as pyc
 
 try:
-    import taco
+    import acis_taco as taco
 except ImportError:
     import Chandra.taco as taco
 import xija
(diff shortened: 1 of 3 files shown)


jzuhone:
@taldcroft @jeanconn there's this PR too, which goes along with it. We should decide how we want to handle this now.

jeanconn:
@taldcroft you are on leave, but are you good with just merging this to master now?

jeanconn:
Actually, I was talking about pushing to master without LRB consideration.

taldcroft:
@jeanconn - Sorry that's what I meant but indeed I worded it very poorly.

xija/gui_fit.py
```diff
@@ -28,7 +28,10 @@
 from Chandra.Time import DateTime
 import pyyaks.context as pyc
 
-import Chandra.taco
+try:
+    import acis_taco as taco
```
jeanconn: I think this is fine, but I am a little confused by the application.  To actually execute this code in gui_fit, would we need to backport/install in the 0.17 gtk skare?

jzuhone: @jeanconn I suspect it's not necessary to backport--the reproducibility aspect of the random numbers is actually turned off in `gui_fit`, so that sort of obviates the main purpose of the new code. 

ACIS ops personnel still use `gui_fit` fairly regularly. I have a Bokeh-based version I'm developing for `xijafit`, but it will take a while before it is ready.

jeanconn: OK.  I can confirm that if we were to run this new xija 3.10 in 0.17 skare, the import looks like it would correctly pick up Chandra.taco .  Not sure about the best ways to run model tests these days to quickly confirm the xija.component.heat bits .

``
In [2]: xija.gui_fit.taco
Out[2]: <module 'Chandra.taco' from '/proj/sot/ska/0.17/arch/x86_64-linux_CentOS-5/lib/python2.7/site-packages/Chandra.taco-0.2-py2.7.egg/Chandra/taco/__init__.pyc'>
In [4]: os.environ['SKA']
Out[4]: '/proj/sot/ska/0.17'
In [6]: xija.__version__
Out[6]: '3.10'
```

taldcroft:
This is OK for merging now.  Actually installing to flight will require the usual regression testing of ACIS review tools and LRB approval.  This will mean no other xija updates can get installed in the meantime.  @jzuhone - are you ready to present this update this week?

jzuhone:
@taldcroft yes, I'll present this week.

PR #51, opened 2018-04-18
Allow for two "hrc_bias" parameters

jzuhone:
The ACIS Ops team has discovered that the single `hrc_bias` parameter which adjusts the expected pitch-dependent solar heating if HRC is in the focal plane is not sufficient to reproduce the temperature behavior for both cases: HRC-I is in the focal plane or HRC-S. Since we spend far more time with HRC-S in the focal plane, this parameter is calibrated well for that case but not for HRC-I.

~~This PR allows a thermal model specification to specify that two bias parameters should be set by setting the `hrc_bias` parameter in  the `SolarHeatHrc` class constructor to a 2-element iterable such as a list, which can be done in the JSON file. This creates two bias parameters, `hrci_bias` and `hrcs_bias`, which then will correspond to the different SIM positions.~~

~~My original thought was to simply change the signature of `SolarHeatHrc` to provide for two parameters, but this breaks backward-compatibility, hence the solution presented here, which while a little awkward supports both types of behavior.~~ 

This PR adds a new class, `SolarHeatHrcOpts`, which has two parameters related to the HRC bias, `hrci_bias` and `hrcs_bias`, which apply when HRC-I or HRC-S is in the focal plane.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -326,6 +326,49 @@ class SolarHeatHrc(SolarHeat):
             self.hrc_mask = self.simz_comp.dvals < 0
         self._dvals[self.hrc_mask] += self.hrc_bias
 
+class SolarHeatHrcOpts(SolarHeat):
+    """Solar heating (pitch and SIM-Z dependent, two parameters for
+    HRC-I and HRC-S)
+
+    :param model: parent model
+    :param node: node which is coupled to solar heat
+    :param simz_comp: SimZ component
+    :param pitch_comp: solar Pitch component
+    :param eclipse_comp: Eclipse component (optional)
+    :param P_pitches: list of pitch values (default=[45, 65, 90, 130, 180])
+    :param Ps: list of solar heating values (default=[1.0, ...])
+    :param dPs: list of delta heating values (default=[0.0, ...])
+    :param var_func: variability function ('exp' | 'linear')
+    :param tau: variability timescale (days)
+    :param ampl: ampl of annual sinusoidal heating variation
+    :param bias: constant offset to all solar heating values
+    :param epoch: reference date at which ``Ps`` values apply
+    :param hrci_bias: solar heating bias when HRC-I is in the focal plane.
+    :param hrcs_bias: solar heating bias when HRC-S is in the focal plane.
+    """
+    def __init__(self, model, node, simz_comp, pitch_comp, eclipse_comp=None,
+                 P_pitches=None, Ps=None, dPs=None, var_func='exp',
+                 tau=1732.0, ampl=0.05, bias=0.0, epoch='2010:001',
+                 hrci_bias=0.0, hrcs_bias=0.0):
+        SolarHeat.__init__(self, model, node, pitch_comp, eclipse_comp,
+                           P_pitches, Ps, dPs, var_func, tau, ampl, bias,
+                           epoch)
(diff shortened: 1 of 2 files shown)


jzuhone:
For some reason I cannot specify reviewers, so ping the following:
@taldcroft 
@jeanconn 
@richardjedgar 
@Gregg140

jeanconn:
I think we need to add you as a collaborator @jzuhone before you can request reviews.

taldcroft:
@jzuhone - Have you confirmed that this works as expected when you try to export to a Python file from gui_fit and then back to JSON?  Probably should be OK, but needs checking.

Another option is just make a new class that has the two HRC bias parameters instead of one.  This will be back compatible in the sense that any existing model spec will still (by definition) run, but you can "promote" the models like the DPA model to use the newer class.  You could even take the opportunity to give it a more suitable name [1].  Another slight advantage is just a more direct API, i.e. not having a fundamentally different program behavior depending on the *shape* of an input.

[1]: enough time has passed that I find it confusing for the ACIS DPA model to use a `SolarHeatHrc` class.  You might just call this new class what it is, `SolarHeatACISwithHrcBias`, or whatever.

taldcroft:
:+1: on cleanup and possibly making it less painful to update component classes.

xija/component/heat.py
```diff
@@ -326,6 +326,49 @@ def dvals_post_hook(self):
...
+    def dvals_post_hook(self):
+        """Apply a bias power offset when SIM-Z is at HRC-S or HRC-I.
+        """
+        if not hasattr(self, 'hrci_mask'):
+            self.hrci_mask = (self.simz_comp.dvals < 0) & \
+                             (self.simz_comp.dvals > -86147)
+        self._dvals[self.hrci_mask] += self.hrci_bias
+        if not hasattr(self, 'hrcs_mask'):
+            self.hrcs_mask = (self.simz_comp.dvals <= -86147) & \
+                             (self.simz_comp.dvals >= -104362)
```
taldcroft: Is this lower bound necessary?  I don't know details here, but was just wondering if it is in theory possible to ever have a SIM-z count < -104362.

taldcroft:
Looks good apart from my one minor question.

jzuhone:
Yeah, the lower bound is overkill. I’ll fix it, thanks for pointing that out.

taldcroft:
Code looks good.  Now just some functional validation for the record here (and for LR eventually), e.g.:

- Make a DPA model which is a clone of the current flight but uses the new class.  Make the HRC-I and HRC-S biases the same as the current single HRC bias and confirm same model prediction.
- Then split the two bias values and confirm "expected changes" (i.e. got hotter or got colder).

jzuhone:
@taldcroft sounds good--but just for clarification...

Do you want this included in the PR/code base in some way?

jzuhone:
Hi @taldcroft, 

This figure encapsulates the tests you suggested above for the DPA model:

1. The overlap of the green and orange lines demonstrates that if we take the model with two bias parameters and set both of them to the same value of the original bias parameter, we get the same model output (this has been verified numerically by me also).
2. The red line reflects what happens when we thaw both bias parameters and fit them to historical data. The increase of the model temperature in the HRC-I position and the decrease in the HRC-S position is what we expect.

hrc_bias

PR #52, opened 2018-07-10
Improve solar heat pitch plot in gui_fit

taldcroft:
Based on the nice idea from @matthewdahmer this includes the 1-year extrapolation of solar heating by pitch.   E.g.

image

### DIFF
--- a/xija/__init__.py
+++ b/xija/__init__.py
@@ -3,7 +3,7 @@ from .model import *
 from .component import *
 from .files import files
 
-__version__ = '3.10'
+__version__ = '3.11'
 
 def test(*args, **kwargs):
     '''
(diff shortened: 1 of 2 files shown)


richardjedgar:
Does this work with the 4-pronged solar heat model for the psmc?

taldcroft:
> Does this work with the 4-pronged solar heat model for the psmc?

Not with this exact fix.  That component has its own `plot` method that would need to be updated.

https://github.com/sot/xija/blob/cc03c7d15dd3bbf8de190b354417c22795e9bfbb/xija/component/heat.py#L579

taldcroft:
Merged this to master so it can be used in fitting now.  Will seek FD approval for installation to flight.

PR #53, opened 2018-07-19
Qt port of gui_fit

taldcroft:


xija/gui_fit.py
```diff
@@ -16,23 +17,37 @@
...
 
 import Chandra.taco
 import xija
-import xija.clogging as clogging   # get rid of this or something
+import sherpa.ui as ui
+
+logging.basicConfig(level=logging.DEBUG)
+logging.debug('Importing gui_fit from {}'.format(__file__))
+
+# Globals for fitting, set in main
```
jzuhone: @taldcroft is there a reason you had to switch to using `global`s here? they are not playing nice with the `multiprocessing` module.

taldcroft: @jzuhone - Humm, I don't remember, sorry.  I browsed through the diffs and nothing rang a bell.  I'm sure there was a reason since otherwise I would not choose to replace working and better-structured code with globals like this...

So who knows if this branch is of much value or if you are better off just starting over from the original Gtk and just using this as a reference.

PR #55, opened 2018-11-01
Step function heating term

jzuhone:
This PR originated in a request from @taldcroft regarding the ACA model. It applies a constant power after a particular moment in time, and zero otherwise. The constant power is a fitted parameter. 

I still need to do some more testing of this on the ACA model, but I'm putting it up here for initial comments.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -904,20 +904,20 @@ class ThermostatHeater(ActiveHeatPower):
             ax.set_title('{}: data (blue)'.format(self.name))
             ax.set_ylabel('Power')
 
-class StepFunctionBias(PrecomputedHeatPower):
+class StepFunctionPower(PrecomputedHeatPower):
     """
     A class that applies a constant temperature shift only 
     after a certain point in time.
     """
-    def __init__(self, model, node, time, bias=0.0):
-        super(StepFunctionBias, self).__init__(model)
+    def __init__(self, model, node, time, P=0.0):
+        super(StepFunctionPower, self).__init__(model)
         self.time = DateTime(time).secs
         self.node = self.model.get_comp(node)
-        self.add_par('bias', bias, min=-10.0, max=10.0)
+        self.add_par('P', P, min=-10.0, max=10.0)
         self.n_mvals = 1
 
     def __str__(self):
-        return 'step_bias__{0}'.format(self.node)
+        return 'step_power__{0}'.format(self.node)
 
     def get_dvals_tlm(self):
         """Return an array of zeros => no activation of the heater.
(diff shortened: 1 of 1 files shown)


jzuhone:
ping @jeanconn also

jzuhone:
Here is a screenshot of the ACA model I fit:
screen shot 2018-11-02 at 12 19 54 pm

jzuhone:
And the JSON model file is in this gist: https://gist.github.com/a5ba50faf33c26a8e273e3768f56dabd

xija/component/heat.py
```diff
@@ -903,3 +903,44 @@ def plot_data__time(self, fig, ax):
...
+
+    def get_dvals_tlm(self):
+        """Return an array of zeros => no activation of the heater.
+        """
+        return np.zeros_like(self.model.times)
+
+    def update(self):
+        """Update the model prediction as a precomputed heat.
+        """
+        self.mvals = self.P*np.ones_like(self.model.times)
```
taldcroft: You can use `np.full_like(self.model.times, fill_value=self.P)` to avoid the extra array creation / multiply.

xija/component/heat.py
```diff
@@ -903,3 +903,44 @@ def plot_data__time(self, fig, ax):
...
+        self.mvals[self.model.times < self.time] = 0.0
```
taldcroft: Since `self.model.times` is sorted, it's going to be faster to use
```
idx0 = np.searchsorted(self.model.times, self.time)
self.mvals[:idx0] = 0.0
```
This should work for cases where `self.time` is either before or after the range of `model.times`, but worth checking.

taldcroft:
The bad data gap should extend to 2018:293:00:00:00.  Before that there is a clear excess from the HRMA being abnormally hot.

xija/__init__.py
```diff
@@ -3,7 +3,7 @@
 from .component import *
 from .files import files
 
-__version__ = '3.11.1'
+__version__ = '3.12.0'
```
taldcroft: The version should be just `3.12` if there is no bug-fix update.

jeanconn:
For Ska flight and LR, are the ska_testr tests enough to show this is a no-op for the non-ACA models, or do we need ACIS to explicitly test?

jzuhone:
There is no need in my opinion for ACIS to test, since the new class will not be brought into our models now and there are no other changes.

jzuhone:
Who should bring this forward for LR approval?

jeanconn:
I can bring the xija change as a "starcheck" change to LR.

PR #63, opened 2019-05-17
Use a pre-computed grid of Earth visibility in the ACIS FP model

### LINKED ISSUE #62: Pre-compute Earth illumination values to speed up ACIS FP model
Conceptually it should be possible to pre-compute the Earth illumination values in order to substantially speed up ACIS FP model evaluation.  There is a question of gridding.  Offhand perhaps it would be sufficient to have:

- 20000 points evenly distributed over the hemisphere where illumination > 0
- 1000 points for Earth distance, probably spaced somewhat logarithmically (as Earth gets farther away then changes make less of a difference).  At some point Earth is effectively point-like and one can just use 1/r^2.

This would result in a data file around 80 Mb assuming 20 million float32 values.

Current performance is faster than 2 ms / point, so this can be computed in less than a day.

[EDIT: number of grid points substantially reduced from this in #63 without impacting accuracy substantially]

taldcroft:
This implements the idea of using a pre-computed grid of Earth visibility (#62) and gives over 200 times speedup of the ACIS FP model.  For a run covering three weeks, the current flight model (computing Earth visibility on-the-fly) takes 27 ms compared to 5.9 sec.

Grid properties:
- 12288 points over the full sphere, so a typical spacing of about 2 deg.  Uses [astropy-healpix](https://astropy-healpix.readthedocs.io/en/latest/index.html) to efficient map points on a sphere to a linear index.
- 100 logarithmically spaced points for the Earth *altitude* from 200 km to 300000 km.  This altitude is on top of the Earth radius of 6371 km used in `acis_taco`.

Validation and performance are in these notebooks:
- [acisfp-grid-validation.ipynb](https://nbviewer.jupyter.org/url/cxc.cfa.harvard.edu/mta/ASPECT/ipynb/xija/acisfp-grid-validation.ipynb)
- [acisfp-grid-performance.ipynb](https://nbviewer.jupyter.org/url/cxc.cfa.harvard.edu/mta/ASPECT/ipynb/xija/acisfp-grid-performance.ipynb)

The grid file is stored in the package as a gzip binary file and is about 1.5 Mb.

The file was computed in about 25 mins on my laptop with:
- https://nbviewer.jupyter.org/url/cxc.cfa.harvard.edu/mta/ASPECT/ipynb/xija/make-vis-grid.ipynb
- https://github.com/sot/acis_taco/pull/13

This PR and the `acis_taco` PR have been reviewed by ACIS and approved.  There are small discrepancies with the current flight output (shown in the histogram below of differences (current flight - new) for 2018:300 to 2019:130):

- Largest diffs are at the lowest temperature when the model is cycling on the heater, and that is not a problem.  
- Diffs of up to 0.1 degC are apparent, but this is well below the model accuracy.

image

image

Closes #62 

Note: this requires a new package astropy-healpix.   We would need to include this in Ska3.


mbaski:
@taldcroft do you envision keeping the data file in that location (...ska/data/acis_taco/)?  If so, it will be another file that will need to be synced (or at least installed) for Windows users.  That's not a big deal, but if it doesn't change very often and could go where the model json file is stored (...\chandra_models\chandra_models\xija\acisfp\), it would essentially come "free" for Windows users.

jzuhone:
@mbaski if it's part of the Python package does it not get installed properly on Windows?

jzuhone:
Wait, never mind. I see what you mean now. But @taldcroft I think we can mitigate this issue if we just put it in the `acis_taco` package?

mbaski:
If the data file might change with the same regularity as the model's json file (i.e. model calibration), then chandra_models package is preferable from a MATLAB Tools perspective since updating the model's json file (and any other files in chandra_models) is less overhead and considered just a characteristics update versus the ACIS Taco package which currently requires an update to the Python area update on Windows (the two are in separate areas and controlled somewhat differently).

taldcroft:
Sure we can move that file into some configured package.  There are three options, xija, acis_taco, and chandra_models.  It isn't totally obvious, and my first thought would have been xija since the `EarthHeat` component now needs it by default (though you can still choose to use the acis_taco calculation).

I didn't put it in to this point because it's at the edge (10 Mb) where I hesitate to put in data files into a github-configured package.  And if we discover that the gridding needs to be reduced then it could grow to 40 Mb.  But maybe chandra_models is a good place since it is a kind of oddball package anyway, and so far my testing shows that the 10 Mb version is good enough.  I think it will zip down a bit more, maybe 7 Mb.  

I also am pretty sure that int16 (0 - 65535) will be enough dynamic range.  In that case you can get down to 1.5 Mb zipped at which point it is getting into the noise.

taldcroft:
Here is the histogram of residuals for 2018:300 to 2019:130.  As mentioned, all of the larger outliers are when the model is on the heater, and so unrelated to this patch:

image

taldcroft:
Here is the same plot using a data grid that has been quantized to 16 bits using:
```
            scale = 65536 / 3.0
            vis_grid = np.round(vis_grid * scale) / scale
```
The largest visibility value in the grid is about 2.73.  I can see tiny differences proving that the change actually did *something*, but from a practical perspective they are equivalent.

image

xija/component/heat.py
```diff
@@ -400,13 +408,58 @@ def __init__(self, model, node,
...
+    def calc_earth_vis_from_grid(self, ephems, q_atts):
+        import astropy_healpix
+        healpix = astropy_healpix.HEALPix(nside=32, order='nested')
+
+        if not hasattr(self, 'earth_vis_grid'):
+            with fits.open(self.earth_vis_grid_path) as hdus:
+                hdu = hdus[0]
+                hdr = hdu.header
+                vis_grid = hdu.data / hdr['scale']  # 12288 x 100
+                self.__class__.earth_vis_grid = vis_grid
```
jzuhone: Could you elaborate on why this is being set this way?

xija/component/heat.py
```diff
@@ -400,13 +408,58 @@ def __init__(self, model, node,
...
+
+                alts = np.logspace(np.log10(hdr['alt_min']),
+                                   np.log10(hdr['alt_max']),
+                                   hdr['n_alt'])
+                self.__class__.log_earth_vis_dists = np.log(hdr['earthrad'] + alts)
```
jzuhone: Same here as above.

xija/component/heat.py
```diff
@@ -384,6 +388,10 @@ def dvals_post_hook(self):
 
 class EarthHeat(PrecomputedHeatPower):
     """Earth heating of ACIS cold radiator (attitude, ephem dependent)"""
+
+    use_earth_vis_grid = True
```
jzuhone: Should this be configurable?

jzuhone:
@taldcroft I had a look at the predictions for the earth solid angle for the flight runs of the last three loads (MAY0619, MAY1319, MAY2019) compared to the run using this functionality. 

The below plots show the runs themselves and the diffs between the runs, where the diff is showing the relative error. The diffs can be as large as 200%, but these are cases where one run has a value of 0 and the other has a non-zero value. I think the agreement is more or less good. 
 
MAY0619 (both runs, blue is flight run, orange is grid run):
may0619_esa
Difference between two runs for MAY0619:
may0619_esa_diff

MAY1319 (both runs, blue is flight run, orange is grid run):
may1319_esa
Difference between two runs for MAY1319:
may1319_esa_diff

MAY2019 (both runs, blue is flight run, orange is grid run):
may2019_esa
Difference between two runs for MAY2019:
may2019_esa_diff

xija/component/heat.py
taldcroft: Here and below, I just needed a place to cache the pre-computed data and grid points so they only get read in once.  Storing as class attributes seems convenient enough won't require reading the data again if you make a new model instance.  The other option I thought of was a module global, but this seemed more contained and localized (and I think this is more well-behaved for multi-processing on Windows, if we ever go that direction).  I also have used things like a global LazyDict that knows how to load data on demand, but that seemed like overkill.  Other ideas?

@mbaski @jskrist - in the MATLAB implementation it is important to recognize this one-time startup cost (about 1 sec in my profiling).  I assume ORviewer is using a persistent Python interpreter session for typical model evaluation?

xija/component/heat.py
taldcroft: Good question.  My feeling was that we do not need to support this up through the level of a model-spec configurable variable.  I.e. I think we are pretty much just going to always run with the fast grid-based model going forward.  Even for model fitting you would want to use the grid model,  both for speed and for fidelity (fitting with the same code that you use for evaluation later).  Thoughts?

For debugging and validation it is already configurable in the sense that you can create a model and then set that class variable on-the-fly to change how the model then gets evaluated.

jzuhone: Makes sense!

taldcroft:
@jzuhone - thanks for making these plots.  I got a little distressed on first glance at the plots "Difference between two runs for ..." because I interpreted them as the absolute difference between Earth visibility in the two cases.  But now I see (please confirm) those are the fractional difference, i.e. something like `(new - old) / old`.

I forgot that the visibility plots are typically shown in log space.  This shows things a little better and highlights the grid and FITS file quantization.  In the MAY0619 run you see around day 131 where the grid model is jumping up and down in discrete steps.  This is the result of not interpolating spatially, instead just using the nearest value.  It would certainly be possible to do the spatial interpolation but will slow things down and I think it will not substantially change the model prediction.

xija/component/heat.py
jskrist: @taldcroft - The current implementation does use a persistent Python session.  The one time startup cost is good to know about, but it won't affect the MATLAB implementation.  When profiling the MATLAB code, I always take an average over a set of runs since the initial run can be affected by this kind of thing, and in practice, users will be running this model many times, so a small hit on the first run is acceptable.

jzuhone:
@taldcroft the diffs are `(new - old) / (0.5*(new+old))`, to avoid some issues when dividing by zero, but that's basically correct.

xija/component/heat.py
```diff
@@ -400,13 +408,58 @@ def __init__(self, model, node,
         self.n_mvals = 1
         self.add_par('k', k, min=0.0, max=2.0)
 
+    def calc_earth_vis_from_grid(self, ephems, q_atts):
+        import astropy_healpix
```
jeanconn: With this as a non-top import, should astropy_healpix be added as a xija package (or conda) requirement or not?

jeanconn:
Does this get hit by any of the xija built-in tests or does it need a new test?

jeanconn:
TWG approved this at the 23-May-2019 meeting.  @taldcroft commented that we did not need a new unit test or regression test to approve and merge but could consider adding one in the future.

PR #64, opened 2019-05-23
WIP: Faster AcisDpaStatePower re-initialization

taldcroft:
This is just playing around with making it possible to more quickly re-compute a model that use `AcisDpaStatePower` in the case of just updating the state values.  The use case is evaluating possible chip-drop scenarios in bulk. 

For a two-week span, this gives about a factor of 8 or 9 speed-up, not counting the time required to set up the new state values.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -821,30 +821,32 @@ class AcisDpaStatePower(PrecomputedHeatPower):
     @property
     def par_idxs(self):
         if not hasattr(self, '_par_idxs'):
-            # Make a regex corresponding to the last bit of each power
-            # parameter name.  E.g. "pow_1xxx" => "1...".
-            power_par_res = [par.name[4:].replace('x', '.')
-                             for par in self.power_pars]
+            if not hasattr(self, '_par_idxs_static'):
+                print('computing new _par_idxs')
+                # Make a regex corresponding to the last bit of each power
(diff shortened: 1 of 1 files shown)


jzuhone:
@taldcroft can you rebase on master to get rid of the Earth heating updates

taldcroft:
@jzuhone - I rebased so there is just the one commit now.

xija/component/heat.py
```diff
@@ -821,30 +821,32 @@ def __str__(self):
...
-                            idx = int(state)
-                            for i, power_par_re in enumerate(power_par_res):
-                                if re.match(power_par_re, state):
-                                    par_idxs[idx] = i
-                                    break
-                            else:
-                                raise ValueError('No match for power state {}'
-                                                 .format(state))
+            if not hasattr(self, '_par_idxs_static'):
+                print('computing new _par_idxs')
```
jzuhone: Unless you think we should leave it in, I'd prefer to get rid of the print statement before merging.

taldcroft: No, I didn't intend to leave that.  In fact this whole PR was a bit of just toying around for a use case that may or may not happen.  But I don't see any harm in this as long as someone else has looked at the code.

taldcroft:
Not planning to merge so closing.

jzuhone:
I might incorporate this in a refactor of this class I am planning.

PR #66, opened 2019-06-18
Add option for multiplicative yearly dependence of solar heating

jzuhone:
The current code for computing the yearly variation in solar heating due to the change in the Earth's distance to the Sun has a bug. The mathematical background which led me to discover this is found here:

[solarheat_orbit.pdf](https://github.com/sot/xija/files/3302107/solarheat_orbit.pdf)

This PR does the following things:

- Fixes the variation in solar heating
- ~~Adds this effect to the `SolarHeatOffNomRoll` and `EarthHeat` classes by creating a `SolarHeatBase` superclass which `SolarHeat` and its subclasses also use.~~
- Adds a new `NewSolarHeat` class so we can implement this fix in models one by one without having to change them all at once. Eventually the `NewSolarHeat` class should replace the `SolarHeat` class. 

Preliminary tests using the ACIS DPA and DEA models show improvements in fitting hot and cold seasons simultaneously. ~~I still want to spend some more time fitting this model and the ACIS FP model to verify that this fix helps.~~ Implementing this into the ACIS FP model requires more work.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -70,7 +70,7 @@ class SolarHeatBase(PrecomputedHeatPower):
     @property
     def h_phase(self):
         if self._h_phase is None:
-            e = 0.0167
+            e = 0.0167 # earth eccentricity
             self._h_phase = (1.0+2.0*e*np.cos(self.t_phase))
         return self._h_phase
 
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -51,28 +51,6 @@ class ActiveHeatPower(ModelComponent):
     pass
 
 
-class SolarHeatBase(PrecomputedHeatPower):
-
-    _t_phase = None
-
-    @property
-    def t_phase(self):
-        if self._t_phase is None:
-            time2000 = DateTime('2000:001:00:00:00').secs
-            time2010 = DateTime('2010:001:00:00:00').secs
-            secs_per_year = (time2010 - time2000) / 10.0
-            t_year = (self.pitch_comp.times - time2000) / secs_per_year
-            self._t_phase = t_year * 2 * np.pi
-        return self._t_phase
-
-    def h_phase(self, ampl=None):
-        if ampl is None:
-            e = 0.0167 # earth eccentricity
-            ampl = 2.0*e
-        h_phase = (1.0+ampl*np.cos(self.t_phase))
-        return h_phase
-
-
 class SolarHeatOffNomRoll(PrecomputedHeatPower):
     """
     Heating of a +Y or -Y face of a spacecraft component due to off-nominal roll.  The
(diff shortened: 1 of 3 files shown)


jzuhone:
ping @taldcroft @matthewdahmer

taldcroft:
I agree a multiplicative factor instead of an additive term generally makes sense.  

The caveat, and part of the reason for the current implementation, is the free bias offset in solar heating (i.e. the degeneracy with the heat sink component).  There is no requirement in the model that solar heat illumination is positive and actually proportional to physical heating.  We've gotten more attentive to this in recent years by trying to make sure eclipse data fit, so in that case the mathematics in the PDF would apply, but for a model with all the P values negative, the additive model would end up working better.

So if we go forward with this (and I'm excited to see performance improvements!), we need to ensure that all flight models are defined such that the P values are generally positive and physically meaningful.

xija/component/heat.py
```diff
@@ -265,6 +287,37 @@ def plot_solar_heat__pitch(self, fig, ax):
             ax.grid()
 
 
+class NewSolarHeat(SolarHeat, SolarHeatBase):
```
jzuhone: I'm having trouble deciding on a new name for this. @taldcroft @matthewdahmer @jeanconn any suggestions? The only "new" thing it would incorporate is the change to the calculation of the yearly variation.

taldcroft: Can you just do this like so:
```
class SolarHeatMultiplicative(SolarHeat):
     # Include the new t_phase property and h_phase method
     # Include the override of dvals
```
When I look at this now it is confusing to add the phase stuff in a new base class that is not used anywhere else (in particular it is not the base for `SolarHeat`).

xija/component/heat.py
```diff
@@ -51,6 +51,28 @@ class ActiveHeatPower(ModelComponent):
     pass
 
 
+class SolarHeatBase(PrecomputedHeatPower):
+
+    _t_phase = None
+
+    @property
+    def t_phase(self):
```
taldcroft: This needs a deleter since there is a `delattr(self, 't_phase')` in the `epoch` setter.  (Don't know why I didn't write that as `del self.t_phase`, but no mind.).

xija/component/heat.py
taldcroft: I didn't check if the new `t_phase` property is exactly the same as what was getting set in `SolarHeat.dvals`, but if so then it could make sense to add the `t_phase` property to `SolarHeat`, modify that `dvals` to use it, and then just have `SolarHeatMultiplicative` add the `h_phase` and override `dvals`.  In fact there should be just one line of difference, so you could factor that statement out to a `_calc_dvals()` method:
```
def _calc_dvals(self, P_vals, dP_vals):
    # This is overridden by the multiplicative version in SolarHeatMultiplicative
    out =(P_vals + dP_vals * self.var_func(self.t_days, self.tau)
             + self.ampl * np.cos(self.t_phase)).reshape(-1)
    return out
```

jeanconn:
Old news, but wouldn't hurt to update the PR top-text with the final names of the classes.

PR #67, opened 2019-06-25
WIP: xija model runner for eval of multiple obs requests

taldcroft:
This is an API stub to provide the basis for discussion on quickly evaluating multiple ORs for thermal constraints from ORviewer during schedule building.

cc: @matthewdahmer @jskrist

### DIFF
--- a/xija/runner.py
+++ b/xija/runner.py
@@ -1,0 +1,212 @@
+# Licensed under a 3-clause BSD style license - see LICENSE.rst
+"""
+Efficiently run multiple xija models on a list of OR's.
+"""
+
+from chandra_aca.transform import calc_aca_from_targ
+
+
+DEFAULT_MODELS = ('aca', 'dpa', 'dea', 'acisfp', 'tcylaft6')  # etc
+
+
+class ModelComp:
+    """
+    Descriptor to specify a model component
+    """
+    def __init__(self, limit=None, model_err=None):
+        self.limit = limit
+        self.model_err = model_err
+
+    def __get__(self, instance, owner_cls):
+        if instance is None:
+            # When called without an instance, return self to allow access
+            # to descriptor attributes.
+            return self
+        else:
+            return getattr(instance, self.name)
+
+    def __set__(self, instance, value):
+        setattr(instance, self.name, value)
+
+    def __set_name__(self, owner_cls, name):
+        self.name = name
+
+        if not hasattr(owner_cls, 'comp_names'):
+            owner_cls.comp_names = []
+        owner_cls.comp_names.append(name)
+
+
+class ModelRunner:
+    """
+    Class object that knows how to run a particular model.  This
+    needs to be subclassed with specifics of the model components
+    that needs to be set and relevant model values that are output.
+    """
+    def run(self, obs_req, start):
+        """
+        Run the model for the specified ``obs_req``
+
+        This sets an attribute ``final_state``
+        """
+        self.states = self.make_states(obs_req)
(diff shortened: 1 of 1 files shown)


xija/runner.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def make_states(self, obs_req, start):
+        """
+        Make state values required to run the model for ``obs_req``.
+
+        Includes:
+
+        - Generate a maneuver to the new attitude
+        - Assume a SIM translation (if necessary) in the middle of the maneuver
+          (is there a better choice?)
```
jskrist: I believe that ORViewer inserts a SIM move into the DOT after the Maneuver statement.

xija/runner.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        pass
+
+    def check_against_limits(self):
+        """
+        Check that the computed temperature(s) are within specified limits.
+
+        This could in theory be a tri-state
+        value: OK, not OK, and almost OK (meaning that the worst case violation
+        was within e.g. 20% of the model error).
```
jskrist: It might be better to return the difference from the model limits and let the end user application determine if that is OK, Not OK, etc.  For example, for a model that is well within the limits, the final application might show a green indicator, but for a model that is near one of the limits, (within some percentage as described in the stub comments here, the raw value is displayed for end user evaluation under a warning yellow/orange indicator.

xija/runner.py
```diff
@@ -0,0 +1,212 @@
...
+
+        :returns: bool (or str?)
+        """
+        pass
+
+
+class AcaModelRunner(ModelRunner):
```
jskrist: I don't fully understand the purpose of this class or how it would be used.  The stub makes it look like it could just as well be a function instead of a class, or a method of the ModelRunner class.

xija/runner.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        self.ephemeris = ephemeris  # needs validation and processing
+        self.obs_reqs = {}
+        self.models = []
+
+        self.model_names = model_names.copy()
+
+        for model_name in model_names:
+            model_runner_name = model_name.capitalize() + 'ModelRunner'
+            self.models[model_name] = globals()[model_runner_name]()
```
jskrist: small syntactical question, are the final parenthesis needed/intended?  I am still building my mental Python parser, but I would think that those parenthesis would cause the variable from `globals()[model_runner_name]` to be evaluated and store the return value, as opposed to storing the value of the variable itself in to the `self.models` dict

xija/runner.py
```diff
@@ -0,0 +1,212 @@
...
+        fep_count, aaccdpt (node value), 1cbat, sim_px (pseudo-node values), sim_z.
+
+        This does assume that comp_names are unique, i.e. there are no models
+        that use the same component name for two different values.  That is
+        currently OK, and should not be a problem in the future.
+
+        IMPLEMENTATION NOTE: can comp_values from MATLAB be a list including
+        different data types?  If not then this method needs to operate on
+        one comp_name/value pair at once.  This is conceptually nicer but
+        probably a bit slower.
```
jskrist: Heterogeneous data can be passed into a python function call from MATLAB using pyexec, although, it may still make more sense to transfer the data into individual variables and then call construct a list/dict/whatever data structure will be used as the inputs in python itself.

taldcroft:
This is not going anywhere, and mostly overtaken by work that @jskrist is doing on the MATLAB side.

PR #68, opened 2019-10-16
Add optional bias parameters for ACIS SolarHeat subclasses for ACIS sim positions

jzuhone:
This PR adds bias parameters for ACIS sim positions to the ACIS-specific `SolarHeat` subclasses, which helps with the ACIS focal plane model. It is designed in such a way so that these parameters must be explicitly set in the JSON model spec file in order to be used, thus ensuring backwards-compatibility.

I also changed the logging level in `gui_fit` to avoid voluminous debug-level messages from Matplotlib on startup.

### DIFF
--- a/xija/__init__.py
+++ b/xija/__init__.py
@@ -3,7 +3,7 @@ from .model import *
 from .component import *
 from .files import files
 
-__version__ = '4.15'
+__version__ = '4.16'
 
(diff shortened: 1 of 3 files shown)


jzuhone:
Declining this PR because upon further review this did not improve model accuracy substantially.

PR #83, opened 2020-04-11
[bugfix] Wrong line is updated in gui_fit

jzuhone:
## Description

This is a quick hotfix for `gui_fit` (I am a bit sheepish about this). During a fit, the data line is being updated in the plot with the model instead of the model line.

This does not affect the model propagation or fitting, and at the end of the fit the model and data lines are correct, but this should fix the model misleadingly looking like a perfect fit while the fit is running. 

## Testing

- [ x] Passes unit tests on MacOS, linux, Windows (at least one required)
- [ x] Functional testing

Fixes #

### DIFF
--- a/xija/component/base.py
+++ b/xija/component/base.py
@@ -266,7 +266,7 @@ class Node(TelemData):
             ax.set_title('{}: model (red) and data (blue)'.format(self.name))
             ax.set_ylabel('Temperature (%s)' % self.units)
         else:
-            lines[1].set_ydata(self.mvals)
+            lines[0].set_ydata(self.mvals)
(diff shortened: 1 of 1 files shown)


jzuhone:
sorry @taldcroft. @Gregg140 this probably explains something you saw that was odd a few weeks back if I remember right.

taldcroft:
Yes that is what I saw but forgot about it.

Gregg140:
Yes you fixed it in the acis py ska version.

jzuhone:
@Gregg140 if I did I must have forgotten about it.

PR #86, opened 2020-06-04
Fix bug in component param set attribute

taldcroft:
## Description

This fixes a 7+ year old bug when trying to do something like:
```
mdl.comp['solarheat__pf0tank2t'].P_60 = 5.0
mdl.calc()
```

Note that a work-around right now is:
```
mdl.comp['solarheat__pf0tank2t'].pars_dict['P_60'] = 5.0
```

## Testing

- [x] Functional testing: New unit testing

### DIFF
--- a/xija/component/base.py
+++ b/xija/component/base.py
@@ -80,7 +80,7 @@ class ModelComponent(object):
             return super(ModelComponent, self).__getattribute__(attr)
 
     def __setattr__(self, attr, val):
-        if attr in self.pars:
+        if attr in self.pars_dict:
             self.pars_dict[attr].val = val
         else:
             super(ModelComponent, self).__setattr__(attr, val)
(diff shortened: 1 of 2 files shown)


taldcroft:
@jskrist - thanks for the detailed testing!  This threw me for a loop for a bit since the new test case I included definitely fails without the patch.  

It turns out that the `heatsink` component uses direct attribute access in the call to `update()` which is used to update based on current parameters.  So in order to get `tau` it does `self.tau`.  The intended design is that `tau` is not a direct object attribute but rather gets found and maintained in the `pars_dict` attribute. However, if you set that object attribute directly as in your test script with the buggy version of xija, it does not get put into `pars_dict` and is just an normal object attribute.

In contrast the test case I used is the solar heat component, and that accesses parameter values with `self.parvals`, which maps into `pars_dict`. So setting `self.P60 = 5` was updating that direct object attribute but not `self.parvals`, hence the problem.

So I learned even a little bit more and feel even better that this is the right change.

PR #88, opened 2020-06-27
SHINY: use kadi command states instead of Chandra.cmd_states

taldcroft:
## Description

Use kadi command states instead of Chandra.cmd_states.

## Testing

- [x] Passes unit tests on MacOS (shiny)
- [x] Expected FAIL tests on MacOS (flight)
- [x] Functional testing (partial)

This uses `kadi.commands.states.interpolate_states` which does not exist in flight. Since the change to using kadi states is slightly controversial, we'll just carry this in the `shiny` branch for now.  However, if the latest `kadi` were installed to flight then these tests should pass.

Ran gui_fit and confirmed that commanded states are read in and the model predition seems reasonable by eye.

### DIFF
--- a/xija/model.py
+++ b/xija/model.py
@@ -18,15 +18,11 @@ import numpy as np
 from . import component
 from . import tmal
 
-try:
-    # Optional packages for model fitting or use on HEAD LAN
-    from Chandra.Time import DateTime, date2secs
-    from astropy.io import ascii
-    import Ska.Numpy
-    import Chandra.cmd_states
-    import Ska.DBI
-except ImportError:
-    pass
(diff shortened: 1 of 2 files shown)


taldcroft:
@jzuhone - merging this to keep the shiny project moving forward, but please review when you have a chance.

jzuhone:
@taldcroft all looks good to me here. I can test if you like.

taldcroft:
@jzuhone - thanks! Hopefully soon I'll be releasing a set of packages comprising the shiny distribution, along with a linux installation on HEAD, GRETA for wider testing. Will be looking for thorough testing then.

PR #96, opened 2020-09-30
Add module to allow getting model spec files from Ska data

taldcroft:
## Description

This adds functions that allow reliably getting xija model spec files from `$SKA/data/chandra_models`. 

By "reliable", the intent is that this will be used in production load review code, replacing the current practice of embedding the spec files into load review code.

The key benefit is allowing more agile xija model updates without requiring code updates and a full Ska3 release. The concept of allowing xija model updates without a code release has been put into practice for MATLAB tools, so extending this to SOT tools seems natural.

Along with this we need the following:
- Clone of the latest flight-approved models git repository in `$SKA/data/chandra_models`
- Process to keep this repo up to date. This can be a manual process that is documented in the Xija model update procedure on the TWG TWiki.  E.g. a Ska admin will keep this checked out at the latest release tag (which by virtue of the model update process guarantees the latest approved flight models).
- Functionality to verify the integrity (repo is clean and at the release tag). This is in the current module.

## Testing

- [x] Functional testing (new tests)

### DIFF
--- a/xija/tests/test_get_model_spec.py
+++ b/xija/tests/test_get_model_spec.py
@@ -7,7 +7,7 @@ import re
 import pytest
 import requests
 
-from ..get_model_spec import (get_xija_model_file, get_xija_model_names,
+from ..get_model_spec import (get_xija_model_spec, get_xija_model_names,
                               get_repo_version, get_github_version)
 
 try:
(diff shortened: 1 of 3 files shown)


xija/tests/test_get_model_spec.py
```diff
@@ -0,0 +1,81 @@
...
+    # Version 3.30
+    spec, version = get_xija_model_spec('aca', version='3.30')
+    assert spec['name'] == 'aacccdpt'
+    assert 'comps' in spec
+    assert spec["datestop"] == "2018:305:11:52:30.816"
+
+
+def test_get_model_spec_aca_latest():
+    # Latest version
+    spec, version = get_xija_model_spec('aca', check_version=HAS_GITHUB)
```
jeanconn, line 31: I'm a little confused here.  I understand the point of checking the capability to check the version, but it seems like the test should pass whether the user has SKA/chandra_models up-to-date (or testing or whatever), as that's not something the module controls.  Seems like the test should make a temporary area for the chandra_models clone and run against that or the like.  Or am I misunderstanding?  Not a big deal either way.

taldcroft: Yes, being pedantic you are right. In practice people should keep their local archive up to date and if they don't then this will fail, but we don't care.  It should always pass on a production / well-maintained machine.

PR #99, opened 2020-10-13
Allow model_spec to be provided as Path

### LINKED ISSUE #98: Allow pathlib Path/PosixPath as model_spec
It looks like:

https://github.com/sot/xija/blob/64b81084e1a4c21c500eb8cd7fdb1d12dc068e73/xija/model.py#L108

only allows model_spec files as string filenames.

taldcroft:
## Description

Fixes #98 and removes some Py2 compatibility, and some whitespace changes tossed in by VS code.

## Testing

- [N/A] Functional testing

Fixes #98

### DIFF
--- a/xija/model.py
+++ b/xija/model.py
@@ -10,8 +10,8 @@ import os
 import json
 import ctypes
 from collections import OrderedDict
-from six.moves import cStringIO as StringIO
-import six
+from io import StringIO
+from pathlib import Path
 
 import numpy as np
 
@@ -61,9 +61,9 @@ class FetchError(Exception):
 class XijaModel(object):
     """Xija model class to encapsulate all ModelComponents and provide the
     infrastructure to define and evaluate models.
-    
+
     The parameters ``name``, ``start``, and ``stop`` are determined as follows:
-    
+
     - If a model specification is provided then that sets the default values
       for keywords that are not supplied to the class init call.
     - ``evolve_method = 1`` uses the original ODE solver which treats every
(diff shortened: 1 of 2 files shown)


xija/model.py
```diff
@@ -103,9 +103,10 @@ class XijaModel(object):
     def __init__(self, name=None, start=None, stop=None, dt=None,
                  model_spec=None, cmd_states=None, evolve_method=None,
                  rk4=None, limits=None):
-        # If model_spec supplied as a string then read model spec as a dict
-        if isinstance(model_spec, six.string_types):
+        # If model_spec is a str or Path then read that file
+        if isinstance(model_spec, (str, Path)):
```
jeanconn, line 107: This makes sense to me.  I can't think of a use case where we'd have something that was recognized as a in six.string_types that wouldn't be fine as a str.  I think my previous question still stands though, are there other classes that would be just as reasonable for a "thing to read for the model spec" as a Path or a str?  And if so, should this just try the read and work by duck typing or does the explicit handling here cover all reasonable cases?  Thanks!

taldcroft: There is always that question. I actually spent some time looking at `open` and the `os.PathLike` protocol. So you are right that a more perfect solution is a try/except after determining that the input model_spec is not None and not a dict (but what if it is a dict-like object that doesn't inherit from dict??).  Except that it ends up being more complex code.

In the end I think this is good enough, and if not we'll fix it then. The user base for this class is quite small.

jeanconn: Thanks for the conversation about it.

jeanconn (approved):

PR #103, opened 2020-10-29
Add MSIDStatePower component

matthewdahmer:
This update adds the ability to apply a heat power shift when a state msid matches a specified value.

## Description

In fitting a new fuel tank model, I found that there is a dependency on the SSR-B state, which can’t be accounted for in Xija without a new heat component. Since I anticipate using this type of state-based heat input for other models in other locations, I built a new flexible Xija heat component that can be defined to use a user-selected msid and state pair.

Here is an example that covers the key information one needs to use this update in the JSON model definition:

In the “comps” section:
```
      {
            "class_name": "MsidStatePower",
            "init_args": [],
            "init_kwargs": {
                "P": 0.5,
                "node": "pf0tank2t",
                "state_msid": "COSSRBX",
                "state_val": "ON "
            },
            "name": "cossrbx_on"
        }
```
In the “pars” section:
```
        {
            "comp_name": "cossrbx_on__pf0tank2t",
            "fmt": "{:.4g}",
            "frozen": false,
            "full_name": "cossrbx_on__pf0tank2t__P",
            "max": 2.0,
            "min": -2.0,
            "name": "P",
            "val": 0.004859426906919793
        }
```

Here is a link to a model that uses this new component: https://github.com/sot/chandra_models/blob/abaa609805431d72da9d4bc949ed5a5c7feb75d4/chandra_models/xija/pftank2t/pftank2t_spec.json

Doc string for this method copied here for convenience:
```
    A class that applies a constant heat power shift only when the state of an
     MSID, ``state_msid``, matches a specified value, ``state_val``.  The shift
     is ``P`` when the ``state_val`` for ``state_msid`` is matched, otherwise it
     is 0.0.

     :param model: parent model object
     :param node: node name or object for which to apply shift
     :param state_msid: state MSID name
     :param state_val: value of ``state_msid`` to be matched
     :param P: size of shift in heat power (default=0.0)

     The name of this component is constructed by concatenating the state msid name
     and the state msid value with an underscore. For example, if the ``MsidStatePower``
     component defines a power value for when the COSSRBX values match the string,
     "ON ", the name of this component is ``cossrbx_on``.

     The ``dvals`` data stored for this component are a boolean type. To initialize
     this component, one would use the following syntax:
         model.comp['cossrbx_on'].set_data(True)
```
## Testing

- [x] Functional testing

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -1065,6 +1065,69 @@ class StepFunctionPower(PrecomputedHeatPower):
             ax.set_ylabel('Power')
 
 
+class MsidStatePower(PrecomputedHeatPower):
+    """
+    A class that applies a constant heat power shift only when the state of an
+    MSID, ``state_msid``, matches a specified value, ``state_val``.  The shift
+    is ``P`` when the ``state_val`` for ``state_msid`` is matched, otherwise it
+    is 0.0.
+
+    :param model: parent model object
+    :param node: node name or object for which to apply shift
+    :param state_msid: state MSID name
+    :param state_val: value of ``state_msid`` to be matched
+    :param P: size of shift in heat power (default=0.0)
+
+    The name of this component is constructed by concatenating the state msid name
+    and the state msid value with an underscore. For example, if the ``MsidStatePower``
+    component defines a power value for when the COSSRBX values match the string,
+    "ON ", the name of this component is ``cossrbx_on``.
+
+    The ``dvals`` data stored for this component are a boolean type. To initialize
+    this component, one would use the following syntax:
+        model.comp['cossrbx_on'].set_data(True)
+
+    """
+    def __init__(self, model, node, state_msid, state_val, P=0.0):
+        super(MsidStatePower, self).__init__(model)
+        self.node = self.model.get_comp(node)
+        self.state_msid = str(state_msid).lower()
+        self.state_val = state_val
+        self.state_val_str = str(state_val).lower().strip()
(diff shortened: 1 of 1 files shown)


taldcroft (approved):
@matthewdahmer - this looks great! I tested it in shiny using your supplied model and I see the substantial improvement.  Great catch on finding that correlation.

My only suggestion was that personally what I wanted to see was just the True/False of the state value matching. Plotting `mvals` was not helpful especially when I played around with the P value and made it zero. Here is code that I used instead. You can choose to add this as another plot option or else replace the `plot_data__time` method. Either way is fine.
```
    def plot_dvals__time(self, fig, ax):
        lines = ax.get_lines()
        if not lines:
            plot_cxctime(self.model.times, self.dvals, '#386cb0', fig=fig, ax=ax)
            ax.grid()
            ax.set_title(f'{self.name}: state match dvals (blue)')
            ax.set_ylabel(f'{self.state_msid.upper()} == {repr(self.state_val)}')
```

matthewdahmer:
@taldcroft I updated the `plot_data__time()` function in the MSIDStatePower component to show the boolean `dvals` data instead of the power values. I re-ran the unit tests on MacOS and re-ran a functional test using xija_gui_fit. Thank you for reviewing this PR!

taldcroft:
Please see the note here: https://github.com/sot/skare3/wiki/Shiny-Ska3-testing#xija regarding promotion of this new functionality.

PR #104, opened 2020-11-12
Use github's 'latest' release link for latest version

jeanconn:
## Description

Instead of fetching all the releases on chandra_models release page, use github's idea of the 'latest' one.

I've already forgotten if we're run into issues with this that could be a problem for chandra_models, but at least it doesn't include pre-releases (which aren't a great conceptual fit with the checks in xija.get_model_spec).

We could probably also filter the old list or some such, but for the given use case, this made sense to me, and maybe it is helpful.

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing (below)

TA: I tried this in my repo after reverting my $SKA/data/chandra_models to 3.32 and it passed tests.


taldcroft (approved):
Very good, I like it!

PR #106, opened 2021-03-03
Add a Delay model component

taldcroft:
## Description

Add a new model component class `Delay` to delay the compute model values from ``node`` by ``delay`` ksec.

For a positive delay, the computed model value (`node.mval`) will be constant at the initial value for the first `delay` ksec. Conversely for a negative delay the values at the end will  be constant for `delay` ksec.

## Testing

- [x] Functional testing

### Functional testing

Made a new thermal model for 4HFSPAT using the `Delay` component (now included in the `examples/delay` directory).
- Confirmed that the model component introduces the expected positive or negative delay, with the endpoints being fixed as expected. 
- Confirmed that the `delay` parameter can be successfully fit with `xija_gui_fit` and that the result is reasonable (i.e. the model and data peaks visually line up).

### DIFF
--- a/examples/delay/4hfspat_delay.json
+++ b/examples/delay/4hfspat_delay.json
@@ -1,0 +1,340 @@
+{
+    "bad_times": [],
+    "comps": [
+        {
+            "class_name": "Node",
+            "init_args": [
+                "4hfspat"
+            ],
+            "init_kwargs": {},
+            "name": "4hfspat"
+        },
+        {
+            "class_name": "Delay",
+            "init_args": [
+                "4hfspat"
+            ],
+            "init_kwargs": {
+                "delay": 0
+            },
+            "name": "delay__4hfspat"
+        },
+        {
+            "class_name": "Pitch",
+            "init_args": [],
+            "init_kwargs": {},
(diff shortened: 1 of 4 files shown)


jzuhone (approved):
Can you remind me what the instability in xija_gui_fit is? And should we fix it?

taldcroft:
@jzuhone  - the instability was just related to this new `Delay` component if the code was
```
comp.node.mvals[:] = np.interp(x=self.times - comp.delay * 1000,
                                                xp=self.times, fp=comp.node.mvals)
```
instead of:
```
comp.node.mvals[1:] = np.interp(x=self.times - comp.delay * 1000,
                                                xp=self.times, fp=comp.node.mvals)[1:]
```
This is what I showed at the TWG when I moved the `delay` slider left and right and the beginning of the model calculation got wacky. This does not happen for the current code using `[1:]` to avoid copying the first element of the array.

matthewdahmer (approved):
I generated two variants of the pline04t model that had a delay and a pseudo node, and one that had the delay with no pseudo node. Both models seemed to run as expected in gui fit.

Issue #105, opened 2021-02-04
xija_gui_fit crashes post-shiny on Linux when updating certain plots


jzuhone:
After the shiny upgrade, `xija_gui_fit` can crash without any error messages on Linux. 

Steps to reproduce:

Run `xija_gui_fit` on the 1DPAMZT model like so:

```
xija_gui_fit --days=365 --stop=2021:018:00:00:00 dpa_model_spec.json
```

Hit the `Histogram` button, and then check the boxes on and off to mask radzones, FMT1, or hit the `Redraw` button. If you do this a few times in close succession, eventually the whole GUI will crash. A similar thing happens if you click and un-click the radzone draw checkbox on the main GUI. 

I've reproduced this both running `xija_gui_fit` on the HEAD LAN and also on a local machine without SSH. @Gregg140 also saw this. 

This potentially has something to do with the button clicking trying to redraw the plot before a previous redraw operation has completed. Oddly, this does not occur on macOS. 

This could potentially be solved by adding threads for the plotting, but I am certainly not an expert on that and the code searches I have done for this seem pretty complex. 

Not clear what changed post-shiny to cause this.

taldcroft:
@jzuhone - I'm not completely clear on what you mean by "changed post-shiny". Do you mean what happened with the big Py 3.8 2020.14 (https://github.com/sot/skare3/pull/545) promotion, or what happened between 2020.14 and now (namely the 2021.1 release)? 

The matplotlib: 2.0.2 -> 3.2.2 update in 2020.14 is an obvious candidate for what changed in 2020.14 that might be causing the issue in `xija_gui_fit`. The 2021.1 release does not contain any package updates that are likely to cause this problem.

jzuhone:
@taldcroft -- what I mean is the big python 3.8 update, and I agree with you that it's probably either the matplotlib (or maybe Qt/pyqt) update that caused the change in behavior.

jzuhone:
Resolved by #108

Issue #70, opened 2019-12-16
Add top-level "limits" key in xija JSON specification


taldcroft:
Based on discussions in recent working groups, and informal support in #69, this issue proposes the formal addition of a top-level keyword `limits` in the xija JSON model specification format.  This key would be a `dict` that can contain the limit keywords shown below.  All keywords would be optional, except that if any values are provided then a `unit` keyword must be supplied.

## Keywords

In general the keywords use the pattern `<limit_system>.<limit_type>.<high/low>.<optional qualifier>`.

### ODB caution and warning limits
- `odb.caution.low`
- `odb.caution.high`
- `odb.warning.low`
- `odb.warning.high`

### Planning limits
- `planning.warning.low` :  "low planning limit" for ACIS DPA for example
- `planning.warning.high` : "planning limit" for most systems (do not exceed)
- `planning.caution.low` : no systems currently use this AFAIK
- `planning.penalty.high` : "penalty limit" for ACA

### Unit for all values
- `unit` : `degF` or `degC`. Defaults to `degC`.

## Examples:

```
# aca_spec.json

    "limits": {
        "aacccdpt": {
            "planning.warning.high": -5.8,
            "planning.penalty.high": -6.8
        }
    },

# dpa_spec.json
        "limits": {
            "odb.caution.high": 39.5,
            "planning.warning.high": 37.5
        },
```

@jzuhone @jeanconn @jskrist @matthewdahmer 

Closes https://github.com/sot/ska-projects/issues/125

taldcroft:
See also the project-level issue https://github.com/sot/ska-projects/issues/125.

jzuhone:
@taldcroft I love this idea. In fact, this was the first thing I tried in #69. I think I just reversed the changes instead of rewriting the history, so I should be able to bring them back easily.

jeanconn:
OK, I had missed this issue.  With regard to using "planning" limits for anything not ODB, are we better off with the capability to just freely label these thins?  That was my suggestion in 

sot/ska-projects#125

But my thought there just speaks to the idea that the MCC API would just be plotting lines for anything not-ODB.  If we have other uses for planning limits, that's a different conversation.

jzuhone:
@taldcroft what color would you like the ACA penalty limit to be?

taldcroft:
> @taldcroft what color would you like the ACA penalty limit to be?

[...]

PR #108, opened 2021-03-18
Fix random crashes using xija_gui_fit on Linux, change limit names for annotation, use CxoTime

jzuhone:
## Description

This PR resolves an issue where `xija_gui_fit` would seemingly randomly crash on Linux when manipulating plots. The exact reason is unknown, because I could never seem to produce a traceback of any kind, but many of the plot objects (such as `Figure` and `Axes` were being re-created every time the data itself needed an update due to a new model fit, etc. This PR implements the changing of the data on the plot itself (for several plot types) instead of re-creating the plot objects with every update, which appears to fix the problem. 

This PR also does the following:

* Uses `CxoTime` instead of `Chandra.Time`
* `annotate_limits` has been moved from being a method of the `XijaModel` class to being a function in the `xija_gui_fit` app, since this is the only place it's used
* Removes old attempt at import of `Chandra.taco` which is a ska2 remnant
* Implement the names for limit values suggested in when plotting them https://github.com/sot/xija/issues/70#issuecomment-780108818

## Testing

- [x] Passes unit tests on macOS
- [x] Functional testing on Linux and macOS (with independent testing from @taldcroft)

Fixes issue #105.

### DIFF
--- a/xija/gui_fit/plots.py
+++ b/xija/gui_fit/plots.py
@@ -419,7 +419,7 @@ class PlotBox(QtWidgets.QVBoxLayout):
 
     def select(self, event):
         grab = event.inaxes and self.main_window.show_line and \
-               self.canvas.toolbar._active is None
+               not self.ax.get_navigate_mode()
         if grab:
             self.mover = self.canvas.mpl_connect('motion_notify_event', self.on_mouse_move)
             self.plots_box.xline = event.xdata
--- a/xija/gui_fit/plots.py
+++ b/xija/gui_fit/plots.py
@@ -160,6 +160,59 @@ def clearLayout(layout):
                 clearLayout(item.layout())
 
 
+def annotate_limits(limits, ax, dir='h'):
+    if len(limits) == 0:
+        return []
+    lines = []
+    draw_line = getattr(ax, f'ax{dir}line')
+    if 'acisi_data_quality' in limits:
+        lines.append(
+            draw_line(limits['acisi_data_quality'],
+                      ls='-.', color='blue')
+        )
+    if 'aciss_data_quality' in limits:
+        lines.append(
+            draw_line(limits['aciss_data_quality'],
+                      ls='-.', color='purple')
+        )
+    if 'planning_caution_high' in limits:
+        lines.append(
+            draw_line(limits['planning_caution_high'],
+                      ls='-.', color='gray')
+        )
+    if 'planning_warning_low' in limits:
+        lines.append(
+            draw_line(limits['planning_warning_low'],
+                      ls='-', color='green')
+        )
+    if 'planning_warning_high' in limits:
+        lines.append(
+            draw_line(limits['planning_warning_high'],
+                      ls='-', color='green')
+        )
+    if 'odb_caution_low' in limits:
+        lines.append(
+            draw_line(limits['odb_caution_low'],
+                      ls='-', color='gold')
+        )
+    if 'odb_caution_high' in limits:
+        lines.append(
+            draw_line(limits['odb_caution_high'],
+                      ls='-', color='gold')
+        )
+    if 'odb_warning_low' in limits:
+        lines.append(
+            draw_line(limits['odb_warning_low'],
+                      ls='-', color='red')
+        )
+    if 'odb_warning_high' in limits:
+        lines.append(
+            draw_line(limits['odb_warning_high'],
+                      ls='-', color='red')
+        )
+    return lines
+
+
 class MplCanvas(FigureCanvas):
     """Ultimately, this is a QWidget (as well as a FigureCanvasAgg, etc.)."""
     def __init__(self, parent=None):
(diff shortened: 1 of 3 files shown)


jzuhone:
I don't see crashes anymore, but I want @Gregg140 to confirm.

xija/gui_fit/plots.py
```diff
@@ -160,6 +160,59 @@ def clearLayout(layout):
                 clearLayout(item.layout())
 
 
+def annotate_limits(limits, ax, dir='h'):
+    if len(limits) == 0:
+        return []
+    lines = []
+    draw_line = getattr(ax, f'ax{dir}line')
+    if 'planning.data_quality.high.acisi' in limits:
```
taldcroft, line 168: These could all be factored into an enclosed function `draw_line(limit_name, linestyle, color)` which has `limits` in scope so you just have a series of calls like:
```
for limit_name, linestyle, color in [
    ('planning.data_quality.high.acisi', '-.', 'blue'),
    (...)]:
    draw_line(limit_name, linestyle, color)
```

xija/gui_fit/plots.py
```diff
@@ -241,12 +294,16 @@ def __init__(self, model, hist_msids):
         self.box.addLayout(toolbar_box)
         self.box.addLayout(check_boxes)
 
-        self.fig = canvas.fig
+        self.limit_lines = []
+
         self.canvas = canvas
+        self.ax1 = self.canvas.fig.add_subplot(121)
```
taldcroft, line 300: Use `1, 2, 1` in new code.

taldcroft (approved):
Just a couple of nitpicky style comments that are optional. I cannot say that I reviewed the logic changes in detail, but I did do a functional test to confirm that the fit method fix worked, and ran `xija_gui_fit` and went through the basic API functionality with no problems. I also ran the unit tests and they pass.

The limit naming looks good, thanks for going along with the suggested naming convention. With this then we can update all the relevant xija models with limits?

Gregg140:
The good news is that I could not make it crash.

I detected a couple of issues in my tests:


1) When I clicked on "Show Limits" for both the main data v time plot and the histogram I never saw the limit lines. Example:

Histo_show_lim


2) After doing some fitting, I asked for the histogram to be re-drawn and I got the following picture.

Pad_Plus_y


The issue is that on the right hand plot I can't tell if the data continues higher than ~13 or it I'm seeing the precise top of the data. A little padding would be good so that I'm sure I'm seeing the top of the histogram.



Example of tests performed:

xija_gui_fit dpa_model_spec.json --days 365 --stop 2021:079:00:00:00.00 &

 1) Delete dpa0 solar heat plot
 
 2) Add 1dpamzt resid_time plot
 
 3) Clicked on show limits
     **- Limits did not appear on the model v data plot**
     
 4) Click Histogram
 
 5) Histogram plot: Show limits
 
     **- Limits did not appear on left hand plot**
     
 6) Mask FMT1
 
 7) Mask Radzones
 
 8) Redraw Histo
  
 9) Histo limits off
 
10) Mask FMT1 off

11) Mask Radzones off

12) fit tau, ampl, bias

13) Mask FMT1, radzones, show limits

14) Unmask FMT!, radzones, show limits

15) Redraw

16) Annotate line

17) Click on main plot to change annotation values

18) Redraw histo

19) Close histo

20) Open histo

21) Redraw histo

22) add resid_data plot

23) Fit P's

24) Redraw Histo

     **Need to pad the +y axis**

jzuhone:
@Gregg140 this should be fixed now.

Gregg140:
Crashed when running through the tests I listed. But I winnowed it down to this:


 1) fit tau, ampl, bias

 2) Click on histogram

Traceback (most recent call last):
  File "/data/acis/ska_pkg/xija/xija/gui_fit/app.py", line 932, in make_histogram
    self.hist_window = HistogramWindow(self.model, self.hist_msids)
  File "/data/acis/ska_pkg/xija/xija/gui_fit/plots.py", line 275, in __init__
    self.make_plots()
  File "/data/acis/ska_pkg/xija/xija/gui_fit/plots.py", line 336, in make_plots
    self.update_plots()
  File "/data/acis/ska_pkg/xija/xija/gui_fit/plots.py", line 416, in update_plots
    self.ax2.set_ylim(0.0, hist.max()+1)
  File "/data/acis/miniconda3/envs/ska-dev/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 3578, in set_ylim
    top = self._validate_converted_limits(top, self.convert_yunits)
  File "/data/acis/miniconda3/envs/ska-dev/lib/python3.8/site-packages/matplotlib/axes/_base.py", line 3101, in _validate_converted_limits
    raise ValueError("Axis limits cannot be NaN or Inf")
ValueError: Axis limits cannot be NaN or Inf

Gregg140 (approved):
Based upon my limited testing of the tool I approve this change.

jzuhone:
@taldcroft this is ready to go.

taldcroft (approved):
LGTM thanks!

PR #109, opened 2021-05-05
This fixes a bug that was missed in the recent gui_fit PR

jzuhone:
## Description

This fixes a bug in `xija_gui_fit` that was missed in the last PR. The bug surfaces when one toggles the limit lines on and off, and causes a crash. 

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing


taldcroft (approved):
LGTM

PR #112, opened 2021-07-20
Two new functions to handle limit strings, minor viz fixes for xija_gui_fit

jzuhone:
## Description

Based on the new limit scheme implemented in https://github.com/sot/chandra_models/pull/75, this PR implements two new functions in xija, `get_limit_spec` and `get_limit_color`. 

`get_limit_spec` takes a limit of the form `<system>.<type>.<direction>.<qualifier>` and returns a dict with these names as the keys:

```python
from xija.limits import get_limit_spec
limit_spec = get_limit_spec("planning.data_quality.high.acisi")
print(limit_spec)
```
which gives
```pycon
{
    'system': 'planning', 
    'type': 'data_quality', 
    'direction': 'high', 
    'qualifier': 'acisi'
}
```
This may have other uses eventually, but it was useful for `get_limit_color`, which uses it to assign colors for the various limits.

`get_limit_color` is used by `xija_gui_fit` to assign limit line colors, but I'll be using and `get_limit_spec` and `get_limit_color` in `acis_thermal_check` and ACISpy. 

Functional testing consisted of running `xija_gui_fit` for every model in `chandra_models` with limits and making sure that the limit lines appeared correctly in the `data__time`, `resid__data`, and histogram plots. 

This PR also fixes a number of minor visualization issues with `xija_gui_fit`. 

## Testing

- [x] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

Fixes #

### DIFF
--- a/xija/gui_fit/plots.py
+++ b/xija/gui_fit/plots.py
@@ -238,6 +238,7 @@ class HistogramWindow(QtWidgets.QMainWindow):
         limits_check.stateChanged.connect(self.plot_limits)
         if len(self.model.limits) == 0:
             limits_check.setEnabled(False)
+        self.limits_check = limits_check
 
         toolbar_box.addWidget(msid_select)
         toolbar_box.addWidget(redraw_button)
@@ -296,21 +297,27 @@ class HistogramWindow(QtWidgets.QMainWindow):
         self.rz_masked = state == QtCore.Qt.Checked
         QtCore.QTimer.singleShot(200, self.update_plots)
 
+    def _clear_limits(self):
+        [line.remove() for line in self.limit_lines]
+        self.limit_lines = []
+
     def plot_limits(self, state):
         if state == QtCore.Qt.Checked:
-            limits = self.model.limits[self.hist_msids[self.which_msid]]
-            self.limit_lines = annotate_limits(limits, self.ax1)
+            limits = self.model.limits.get(self.hist_msids[self.which_msid], None)
+            if limits is not None:
+                self.limit_lines = annotate_limits(limits, self.ax1)
         else:
-            [line.remove() for line in self.limit_lines]
-            self.limit_lines = []
+            self._clear_limits()
         self.canvas.draw_idle()
 
     def change_msid(self, msid):
+        self._clear_limits()
         self.which_msid = self.hist_msids.index(msid)
         self.comp = self.model.comp[self.hist_msids[self.which_msid]]
         msid_name = self.hist_msids[self.which_msid]
         self.ax1.set_title(f'{msid_name}: data vs. residuals (data - model)')
         QtCore.QTimer.singleShot(200, self.update_plots)
+        self.plot_limits(self.limits_check.checkState())
 
     def make_plots(self):
         msid_name = self.hist_msids[self.which_msid]
@@ -433,7 +440,7 @@ class HistogramWindow(QtWidgets.QMainWindow):
             self.plot_dict["max_text"] = self.ax2.text(
                 xpos_max, ystart, 'Maximum Error', 
                 ha="left", va="center", rotation=90)
-        
+
         self.canvas.draw_idle()
 
 
--- a/xija/gui_fit/plots.py
+++ b/xija/gui_fit/plots.py
@@ -368,9 +368,14 @@ class HistogramWindow(QtWidgets.QMainWindow):
         min_resid = np.nanmin(resids)
         max_resid = np.nanmax(resids)
 
+        dvalsr = dvals + randx
+        
+        min_dvals = np.nanmin(dvalsr)
+        max_dvals = np.nanmax(dvalsr)
+
         if len(self.plot_dict) == 0:
             self.plot_dict['resids'] = self.ax1.plot(
-                resids, dvals + randx, 'o', color='#386cb0',
+                resids, dvalsr, 'o', color='#386cb0',
                 alpha=1, markersize=1, markeredgecolor='#386cb0')[0]
             self.plot_dict["01"] = self.ax1.plot(
                 Epoints01, Tpoints01, color='k', linewidth=4)[0]
@@ -396,7 +401,7 @@ class HistogramWindow(QtWidgets.QMainWindow):
                 max_resid, color='k', linestyle='--', 
                 linewidth=1.5, alpha=1)
         else:
-            self.plot_dict['resids'].set_data(resids, dvals + randx)
+            self.plot_dict['resids'].set_data(resids, dvalsr)
             self.plot_dict['01'].set_data(Epoints01, Tpoints01)
             self.plot_dict['99'].set_data(Epoints99, Tpoints99)
             self.plot_dict['50'].set_data(Epoints50, Tpoints50)
@@ -408,6 +413,9 @@ class HistogramWindow(QtWidgets.QMainWindow):
             self.plot_dict['min_hist'].set_xdata(min_resid)
             self.plot_dict['max_hist'].set_xdata(max_resid)
             self.plot_dict['fill'].remove()
+
+        self.ax1.set_xlim(min_resid-0.5, max_resid+0.5)
+        self.ax1.set_ylim(min_dvals-0.5, max_dvals+0.5)
 
         self.plot_dict["fill"] = self.ax2.fill_between(
             bin_mid, hist, step="mid", color='#386cb0')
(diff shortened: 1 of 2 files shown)


jzuhone:
@taldcroft Wanted to see if you had any comments on this.

xija/limits.py
```diff
@@ -0,0 +1,60 @@
...
+    Based on an input *limit* string of the form:
+
+    <system>.<type>.<direction>.<qualifier>
+
+    (where the "qualifier" element is optional) return the various pieces 
+    into a dictionary for convenience.
+    """
+    words = limit.split(".")
+    if len(words) < 3:
+        raise RuntimeError(f"{limit} is not a valid limit string!")
```
taldcroft, line 12: No exclamation point. I mostly follow the Python builtin exceptions and start with lower case and do not end with any punctuation.

xija/limits.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Based on an input *limit* string of the form:
+    
+    <system>.<type>.<direction>.<qualifier>    
+    
+    (where the "qualifier" element is optional) return the color to
+    be used for plotting.
+    """
+    limit_spec = get_limit_spec(limit)
+    if limit_spec["qualifier"] is not None:
```
taldcroft, line 35: I think this logic assumes anything with a qualifier is ACIS. The logic here is bound to get messy, so I think it might be easier to just enumerate the colors in an ordered dict and use wildcards. Something like:
```
LIMIT_COLORS = {
'*.acisi': 'blue',  # could be more specific if that makes sense
...   # and so on
'odb.caution.*': 'gold',
'safety.caution.*': 'gold',
'odb.warning.*': 'red',
...
'planning.caution.*': 'dodgerblue',
...
'*': 'black'   # ?? Either have a catchall or else issue a warning (but not exception)
}
```
This way it is easy to expand, make special cases, and so on. The first match would win.

xija/limits.py
```diff
@@ -0,0 +1,60 @@
...
+        }[limit_spec["type"]]
+    elif limit_spec["system"] == "planning":
+        # Planning limits are green, dodgerblue, or gray 
+        color = {
+            "caution": "dodgerblue",
+            "warning": "green",
+            "penalty": "gray"
+        }[limit_spec["type"]]
+    else:
+        raise RuntimeError(f"limit \"{limit}\" is unknown!")
```
taldcroft, line 59: At worst this should be a warning. You don't want to prevent things from running because of an unknown limit spec.

xija/gui_fit/plots.py
```diff
@@ -161,27 +161,17 @@ def clearLayout(layout):
...
-    for (limit_name, ls, color) in opts:
-        if limit_name in limits:
-            lines.append(
-                draw_line(limits[limit_name], ls=ls, color=color)
-            )
+    for limit in limits:
+        if limit == "unit":
+            continue
+        lines.append(
+            draw_line(limits[limit], color=get_limit_color(limit))
```
taldcroft, line 173: I think this should pay attention to the limit units (indeed many of them are not degC).

jzuhone: @taldcroft what do you want it to do exactly? This loop is to draw the lines and since the `"unit"` item is in the same dict as the limit definitions, it needs to skip over it which is the reason for lines 170-171.

xija/limits.py
jzuhone: I implemented something more or less identical to this. See new version.

xija/limits.py
jzuhone: Fixed.

xija/limits.py
jzuhone: There is no warning or error here anymore.

jzuhone:
This is ready for review again.

xija/limits.py
```diff
@@ -0,0 +1,54 @@
...
+    
+    <system>.<type>.<direction>.<qualifier>    
+    
+    (where the "qualifier" element is optional) return the color to
+    be used for plotting.
+    """
+    import re
+    color = "black"
+    for k, v in limit_colors.items():
+       if re.search(k, limit) is not None:
```
taldcroft, line 52: This should be `fnmatch.fnmatch` instead of `re.search`.  The idea is to use the simpler wildcard matching, otherwise you would need to escape all the periods (`r'planning\.caution\..*'`) and it just looks messy.

xija/limits.py
```diff
@@ -0,0 +1,54 @@
...
+        "direction": words[2]
+    }
+    if len(words) == 4:
+        limit_spec["qualifier"] = words[3]
+    else:
+        limit_spec["qualifier"] = None
+    return limit_spec
+
+
+limit_colors = {
```
taldcroft, line 25: Should be SCREAMING_CAP_CASE as a module constant.

xija/limits.py
```diff
@@ -0,0 +1,54 @@
...
+    <system>.<type>.<direction>.<qualifier>    
+    
+    (where the "qualifier" element is optional) return the color to
+    be used for plotting.
+    """
+    import re
+    color = "black"
+    for k, v in limit_colors.items():
+       if re.search(k, limit) is not None:
+            color = v
```
taldcroft, line 53: I had been thinking that the first match would be returned, whereas this returns the last match. I suspect that most humans naturally expect the first match, but whatever you decide is fine as long as it is documented.

xija/gui_fit/plots.py
taldcroft: E.g the limits for the PFTANK2T model are in degF, but the models are always computed and calculated in degC. Maybe I'm missing the code, but somewhere along the way the limit values need to be converted degC.

xija/gui_fit/plots.py
```diff
@@ -161,27 +161,17 @@ def clearLayout(layout):
 
 
 def annotate_limits(limits, ax, dir='h'):
+    from xija.limits import get_limit_color
```
taldcroft, line 164: In general I think we should reserve local imports for dependencies that are optional to the code running or where import time is a priority for a heavy dependency like pandas. Same below with the local import of `re`, which is basically always imported already by the time you get to this level.

xija/limits.py
```diff
@@ -0,0 +1,54 @@
+def get_limit_spec(limit):
```
taldcroft, line 1: Needs a license line.

taldcroft: This function is not called anywhere here. Is there an idea that it would be useful in some other context? I don't mind leaving it, just asking.

xija/gui_fit/plots.py
```diff
@@ -161,27 +161,17 @@ def clearLayout(layout):
 
 
 def annotate_limits(limits, ax, dir='h'):
```
taldcroft, line 163: Can you add a docstring? Also for the new functions later can you add `:param x: blah` and `returns: blah` for the public funcions?

xija/limits.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Based on an input *limit* string of the form:
+    
+    <system>.<type>.<direction>.<qualifier>    
+    
+    (where the "qualifier" element is optional) return the color to
+    be used for plotting.
+    """
+    import re
+    color = "black"
```
taldcroft, line 50: The default color should be embedded in `LIMIT_COLORS` as `'*': black`.

xija/limits.py
```diff
@@ -0,0 +1,54 @@
...
+    }
+    if len(words) == 4:
+        limit_spec["qualifier"] = words[3]
+    else:
+        limit_spec["qualifier"] = None
+    return limit_spec
+
+
+limit_colors = {
+    '.*.acisi': 'blue',
```
taldcroft, line 26: Should these be `*.*.acisi` to allow working with `get_limit_spec`?

xija/limits.py
jzuhone: Sorry, I should have added a `break` here.

xija/limits.py
jzuhone: Done

xija/gui_fit/plots.py
jzuhone: Done.

xija/limits.py
jzuhone: Added the license line. I originally had `get_limit_spec` used in `get_limit_color`, but now it is not anymore. I think it may be useful later.

xija/gui_fit/plots.py
jzuhone: Done

xija/limits.py
jzuhone: I don't think that this dict is meant to work with `get_limit_spec`.

taldcroft (approved):
Looks good now!

PR #113, opened 2021-08-25
Fixing bugs in limits implementation and adding them to the model info window in xija_gui_fit

jzuhone:
## Description

Hi @taldcroft, 

I should have noted after I made those pushes to #112 last night that I wasn't finished testing, my apologies. There were bugs in the implementation that are now fixed here. 

I also found it useful to add the limits to the "Model Info" window in `xija_gui_fit`--see two screenshots below. If the units are degF, the window also reports the degC values since those are used in the GUI. 

Screen Shot 2021-08-25 at 10 33 50 AM
Screen Shot 2021-08-25 at 10 10 10 AM

## Testing

- [ ] Passes unit tests on MacOS, linux, Windows (at least one required)
- [x] Functional testing

Fixes #

### DIFF
--- a/xija/gui_fit/app.py
+++ b/xija/gui_fit/app.py
@@ -243,6 +243,18 @@ class ModelInfoWindow(QtWidgets.QMainWindow):
         main_box.addWidget(QtWidgets.QLabel("Timestep: {:.1f} s".format(model.dt)))
         main_box.addWidget(QtWidgets.QLabel("Evolve Method: Core {}".format(model.evolve_method)))
         main_box.addWidget(QtWidgets.QLabel("Runge-Kutta Order: {}".format(4 if model.rk4 else 2)))
+        for msid in self.model.limits:
+            limits = self.model.limits[msid]
+            units = limits["unit"]
+            main_box.addWidget(QtWidgets.QLabel(f"{msid.upper()} limits:"))
+        for limit, val in limits.items():
+            if limit == "unit":
+                continue
+            limit_str = f"    {limit}: {val:.1f} {units}"
+            if units == "degF":
+                valC = (val - 32.0)*5.0/9.0
+                limit_str += f" ({valC:.1f} degC)"
+            main_box.addWidget(QtWidgets.QLabel(limit_str))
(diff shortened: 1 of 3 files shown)


xija/gui_fit/app.py
```diff
@@ -243,6 +243,18 @@ def __init__(self, model, main_window):
         main_box.addWidget(QtWidgets.QLabel("Timestep: {:.1f} s".format(model.dt)))
         main_box.addWidget(QtWidgets.QLabel("Evolve Method: Core {}".format(model.evolve_method)))
         main_box.addWidget(QtWidgets.QLabel("Runge-Kutta Order: {}".format(4 if model.rk4 else 2)))
+        for msid in self.model.limits:
+            limits = self.model.limits[msid]
+            units = limits["unit"]
+            main_box.addWidget(QtWidgets.QLabel(f"{msid.upper()} limits:"))
+        for limit, val in limits.items():
```
taldcroft, line 250: Shouldn't this loop be in the `for msid ..` block? Otherwise the `limits` and `units` are just the ones for the last `msid`. I think in practice with the current xija spec files this won't make a difference, but still it doesn't look right to me.

xija/gui_fit/app.py
```diff
@@ -243,6 +243,18 @@ def __init__(self, model, main_window):
...
+            if limit == "unit":
+                continue
+            limit_str = f"    {limit}: {val:.1f} {units}"
+            if units == "degF":
+                valC = (val - 32.0)*5.0/9.0
```
taldcroft: BTW, you can consolidate the temperature conversions (in 2 places) using `cheta.units.F_to_C`. The `cheta.units` import is lightweight.

jzuhone: Thanks! Done

taldcroft:
@jzuhone - the code looks fine to me now, but please do run the unit tests and check that box for the record, then I'll approve.

taldcroft (approved):
@jzuhone - I gave you commit access here. You can merge when ready, and in the future I'll approve but let you merge.

PR #117, opened 2022-03-02
Adjust earthshine effect based on Earth phase

jzuhone:
## Description

NOTE: The PR is finished but there is no expected hurry to merge it since I anticipate it will generate some discussion. 

This PR allows the heating on the ACIS FP from the earthshine into the ACIS radiator to depend on the Earth phase. The existing code assumes that the relevant earthshine is in the infrared, which should to zeroth order be the same regardless if the Earth is illuminated or not by the Sun. This code is an experimental feature to test the hypothesis that the model for the effect of earthshine on the temperature could be improved if we accounted for the Earth phase. 

This PR adds this capability by using the solar ephemeris along with the spacecraft ephemeris to compute the fraction of the  Earth which is illuminated by the Sun. It is currently an optional calculation through the use of kwargs. Note that this is a simple calculation which assumes that the ACIS radiator can see the whole Earth. 

This calculation improves the fit of the ACIS FP thermal model at high temperatures during radzones, which was demonstrated [at the TWG on 3/3/22](https://occweb.cfa.harvard.edu/twiki/bin/view/ThermalWG/MeetingNotes2022x03x03).

## Interface impacts

If the Earth phase factor is not included, the same model results are obtained as with the previous xija version.

### Unit tests

- [x] Mac
### Functional tests

Models with and without the Earth phase factor were checked to ensure they ran as expected.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -659,7 +659,9 @@ class EarthHeat(PrecomputedHeatPower):
 
             # This next bit optionally checks to see if the solar ephemeris
             # was passed in, and if it was it computes the fraction of the
-            # Earth's surface that is illuminated by the Sun.
+            # Earth's surface that is illuminated by the Sun. Originally
+            # discussed at:
+            # https://occweb.cfa.harvard.edu/twiki/bin/view/ThermalWG/MeetingNotes2022x03x03
             solar_xyzs = [getattr(self, f'solarephem0_{x}') for x in 'xyz']
 
             if self.use_earth_phase:
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -544,21 +544,31 @@ class EarthHeat(PrecomputedHeatPower):
 
     def __init__(self, model, node,
                  orbitephem0_x, orbitephem0_y, orbitephem0_z,
-                 solarephem0_x, solarephem0_y, solarephem0_z,
                  aoattqt1, aoattqt2, aoattqt3, aoattqt4,
-                 k=1.0):
+                 solarephem0_x=None, solarephem0_y=None, 
+                 solarephem0_z=None, k=1.0):
         ModelComponent.__init__(self, model)
         self.node = self.model.get_comp(node)
         self.orbitephem0_x = self.model.get_comp(orbitephem0_x)
         self.orbitephem0_y = self.model.get_comp(orbitephem0_y)
         self.orbitephem0_z = self.model.get_comp(orbitephem0_z)
-        self.solarephem0_x = self.model.get_comp(solarephem0_x)
-        self.solarephem0_y = self.model.get_comp(solarephem0_y)
-        self.solarephem0_z = self.model.get_comp(solarephem0_z)
         self.aoattqt1 = self.model.get_comp(aoattqt1)
         self.aoattqt2 = self.model.get_comp(aoattqt2)
         self.aoattqt3 = self.model.get_comp(aoattqt3)
         self.aoattqt4 = self.model.get_comp(aoattqt4)
+        if solarephem0_x is None:
+            self.solarephem0_x = None
+        else:
+            self.solarephem0_x = self.model.get_comp(solarephem0_x)
+        if solarephem0_y is None:
+            self.solarephem0_y = None
+        else:
+            self.solarephem0_y = self.model.get_comp(solarephem0_y)
+        if solarephem0_z is None:
+            self.solarephem0_z = None
+        else:
+            self.solarephem0_z = self.model.get_comp(solarephem0_z)
+        
         self.n_mvals = 1
         self.add_par('k', k, min=0.0, max=2.0)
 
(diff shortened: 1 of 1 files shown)


taldcroft:
> This could also allow models to have a telemetry format dependence in the future (for fitting, most likely) if it were ever desired.

I am having trouble seeing how telemetry format could play into a thermal model, but maybe I'm missing something. There are likely some coincidental correlations with format, but not causation right?

Since this is completely unrelated it would probably be best to put 555c291 into a separate PR.

xija/component/heat.py
```diff
@@ -640,6 +654,21 @@ def dvals(self):
             else:
                 self.calc_earth_vis_from_taco(ephems, q_atts)
 
+            # This next bit optionally checks to see if the solar ephemeris
+            # was passed in, and if it was it computes the fraction of the
+            # Earth's surface that is illuminated by the Sun. 
+            solar_xyzs = [getattr(self, 'solarephem0_{}'.format(x))
+                          for x in ('x', 'y', 'z')]
+
+            if not None in solar_xyzs:
```
taldcroft, line 663: Linters will probably say this should be `if None not in solar_xyzs:`. I'm try to get away from nitpicky comments, but here there is real ambiguity about precedence `not (None in solar_xyzs)` or `(not None) in solar_xyzs`. It is probably the former since the code is working, but this could be worth fixing.

xija/component/heat.py
```diff
@@ -640,6 +654,21 @@ def dvals(self):
             else:
                 self.calc_earth_vis_from_taco(ephems, q_atts)
 
+            # This next bit optionally checks to see if the solar ephemeris
+            # was passed in, and if it was it computes the fraction of the
+            # Earth's surface that is illuminated by the Sun.
```
taldcroft, line 659: Include a link or reference to the TWG meeting where you presented this, so we can get back to that nice hand-drawn picture of the math.

xija/component/heat.py
```diff
@@ -640,6 +654,21 @@ def dvals(self):
...
+                          for x in ('x', 'y', 'z')]
+
+            if not None in solar_xyzs:
+
+                solars = np.array([x.dvals for x in solar_xyzs]).transpose().copy()
+
+                cos = np.sum(ephems*solars, axis=1)
+                es = np.sum(ephems*ephems, axis=1)*np.sum(solars*solars, axis=1)
+                cos /= np.sqrt(es)
+                self._dvals *= 0.5*(1.0+cos)
```
taldcroft, line 670: Should there be a convenient way to get access to this multiplicative factor via plotting or at least as an attribute to allow for diagnostics? I don't remember enough of the details to know how hard this would be.

taldcroft (commented):
The code here looks basically fine, with some minor comments. The PR itself needs docs on the testing, interface impacts (e.g. making sure it doesn't change existing results for the case without solar ephemeris) etc.

Is there benefit to getting this into the upcoming Ska release?

taldcroft:
OK, now I understand about the TlmFormat. So just go ahead with a separate PR.

xija/component/heat.py
jzuhone: Fixed this

taldcroft (approved):
I ran the unit tests on Mac with a pass. Thanks for the changes, looks good now.

matthewdahmer:
I tested this PR on chimchim with a local Xija installation by generating the ACIS FP dashboard plot with your new acisfp_spec.json (https://github.com/jzuhone/chandra_models/blob/fp_recal_2022/chandra_models/xija/acisfp/acisfp_spec.json). Everything seemed to work.

jzuhone:
@taldcroft I will add the extra info to the PR description. Should I bundle this with the new thermal model in a single FSDS?

jzuhone:
@taldcroft @matthewdahmer one note--with the added parameter `k2` that I have now included, the current flight ACIS FP model no longer works with this proposed version of xija, since it does not have the required number of parameters. I suppose this is fine given that we'd want to promote both this version and the new ACISFP model in https://github.com/sot/chandra_models/pull/80, but I wanted you to be aware.

matthewdahmer:
@jzuhone @taldcroft, If I remember correctly, the new ACIS FP model will require changes to the Matlab FOT Tools to pass in new parameters. We should get that sorted out before submitting the FSDS for the thermal model. The new version of Xija should be promoted first and included in the FOT MATLAB Tools version of Ska so @jskrist has the supporting Xija code available to develop against.

jzuhone:
> The new version of Xija should be promoted first and included in the FOT MATLAB Tools version of Ska so @jskrist has the supporting Xija code available to develop against.

But if we promote this version of xija and install it without the new model, we cannot run the existing ACIS FP flight model, correct? It does not have the new parameter.

matthewdahmer:
@jzuhone Sorry, I read what you wrote too quickly. I thought I heard you say during the TWG meeting that the current version of the ACIS FP model would work with the new version of Xija, however I see you are saying this is not the case. 

We just need to:

1. Make sure this Xija update does not make it into the FOT Matlab Tools Ska before the Matlab code is updated.
2. Avoid merging the new ACIS FP model into the main/master branch of chandra_models before we are ready to use it for load reviews so we don't need to reverse-merge it if we have another model we find we need to promote first.

We should tag up with @jskrist during our follow-on meeting after the MPCWG to see what his schedule looks like and how he would prefer we proceed. I'll bring this up during our Matlab Dev meeting tomorrow as well.

jeanconn:
I also note that doesn't seem quite clear from the interface impacts?  If I understand correctly, should this PR be updated to be more backward compatible and then make the breaking change at the next version?

matthewdahmer:
This PR was included in FSDS-25 and is now approved.

jeanconn:
Right, so since the FSDS was brought as a github PR, I think this can get merged and a new release cut.

jeanconn (approved):

PR #120, opened 2022-04-25
Radiative coupling

matthewdahmer:
## Description
This PR adds a radiative coupling to complement the linear conductive coupling.

## Interface impacts

This PR does not result in any backwards incompatible changes, and does not directly impact the use of any models using the radiative coupling within the FOT MATLAB Tools.

## Testing

This was developed and tested within the Mac OS environment. No particular setup is needed for functional testing, outside of a model that includes the radiative coupling component.

### Unit tests

- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

Functional testing to be discussed at the April 28 Thermal Working Group meeting.

### DIFF
--- a/xija/core_1.c
+++ b/xija/core_1.c
@@ -19,7 +19,6 @@ int calc_model_1(int n_times, int n_preds, int n_tmals, double dt,
     double k2, dt2, tempk1, tempk2;
     int i, j, j0, i1, i2, i3, rki, opcode;
 
-    double sb_const = 5.670374419E-8;
     double c2k = 273.15;
 
     for (j0 = 0; j0 < n_times-2; j0 += 2) {
@@ -80,19 +79,15 @@ int calc_model_1(int n_times, int n_preds, int n_tmals, double dt,
                         }
                         break;
                     case 5:  /* Node to node radiative coupling */
+                        tempk1 = (y[i1] + c2k) / 300;
                         if (i2 < n_preds && i1 < n_preds) {
-                            tempk1 = y[i1] + c2k;
-                            tempk2 = y[i2] + c2k;
-                            deriv[i1] += (tempk2 * tempk2 * tempk2 * tempk2 - tempk1 * tempk1 * tempk1 * tempk1)
-                                * tmal_floats[i][0] * sb_const;
-
+                            tempk2 = (y[i2] + c2k) / 300;
                         }
                         else {
-                            tempk1 = y[i1] + c2k;
-                            tempk2 = mvals[i2][j] + c2k;
-                            deriv[i1] += (tempk2 * tempk2 * tempk2 * tempk2 - tempk1 * tempk1 * tempk1 * tempk1)
-                                * tmal_floats[i][0] * sb_const;
+                            tempk2 = (mvals[i2][j] + c2k) / 300;
                         }
+                        deriv[i1] += (pow(tempk2, tmal_floats[i][1]) -
+                            pow(tempk1, tmal_floats[i][1])) * tmal_floats[i][0];
                         break;
                     }
             }
--- a/xija/core_1.c
+++ b/xija/core_1.c
@@ -16,8 +16,11 @@ int calc_model_1(int n_times, int n_preds, int n_tmals, double dt,
 {
     double deriv[100];  /* n_preds -- learn to malloc! */
     double y[100];
-    double k2, dt2;
-    int i, j, j0, i1, i2, i3, rki, opcode;    
+    double k2, dt2, tempk1, tempk2;
+    int i, j, j0, i1, i2, i3, rki, opcode;
+
+    double sb_const = 5.670374419E-8;
+    double c2k = 273.15;
 
     for (j0 = 0; j0 < n_times-2; j0 += 2) {
         for (rki = 0; rki < 2; rki++) {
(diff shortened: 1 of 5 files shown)


taldcroft:
@matthewdahmer - what is the driver for this? I think that in practice (over the typical temperature ranges we see) the T^4 curve is reasonably close to linear. Maybe the proof is in the pudding if you can get a model fit with this new component that is much better than what is possible with the linear conductive coupling.

Here is T^4 in blue and linear in orange.

image

matthewdahmer:
The LAE domes have a direct view to the MUPS valve housings, and since these domes likely heat quickly and likely get rather hot, they may provide a significant radiative heat load to the valves. Interestingly enough, I found that introducing a radiative coupling doesn't have as much of an effect in improving temperature predictions at the hottest valve temperatures as it does at the colder valve temperatures. In the end, this will serve to make momentum unload predictions more accurate.

taldcroft:
@matthewdahmer - my point is that you should try a linear coupling as well to see if using the new T^4 coupling actually makes a noticeable improvement in the prediction compared to the linear version.

About units, xija is basically unit-free, so there is not really a point in introducing `Sb`, all you really need is to make it so that the final parameter that gets fitted is of order a few, which helps out sherpa in fitting. Instead of using the 10-digit `sb_const` you could just use 1e-8 or 1e-9 as an arbitrary scaling.

xija/core_2.c
```diff
@@ -79,6 +82,19 @@ void dTdt(int j, int half, int n_preds, int n_tmals, int **tmal_ints,
                 }
                 if (half == 0) mvals[i3][j] = mvals_i3;
                 break;
+            case 5:  /* Node to node radiative coupling */
+                tempk1 = y[i1] + c2k;
+                if (i2 < n_preds && i1 < n_preds) {
+                    tempk2 = y[i2] + c2k;
+                    deriv[i1] += (tempk2 * tempk2 * tempk2 * tempk2 - tempk1 * tempk1 * tempk1 * tempk1)
```
taldcroft, line 89: Same, factor this out.

xija/core_1.c
```diff
@@ -76,6 +79,21 @@ int calc_model_1(int n_times, int n_preds, int n_tmals, double dt,
                             mvals[i3][j] = 0.0;
                         }
                         break;
+                    case 5:  /* Node to node radiative coupling */
+                        if (i2 < n_preds && i1 < n_preds) {
+                            tempk1 = y[i1] + c2k;
```
taldcroft, line 84: This is where you could make this unitless and of order 1 and get rid of the `sb_const`.
```suggestion
                            tempk1 = (y[i1] + c2k) / 300;
```

taldcroft:
@matthewdahmer - have a look at https://gist.github.com/taldcroft/8f80ab919b211e2c4e96276f2c744aec. This defines a 2nd-order approximation that is worth exploring to see if it improves stability of fitting.

xija/core_1.c
```diff
@@ -76,6 +79,21 @@ int calc_model_1(int n_times, int n_preds, int n_tmals, double dt,
...
+                            tempk2 = y[i2] + c2k;
```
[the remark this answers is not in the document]
matthewdahmer: Done, thanks!

xija/core_1.c
matthewdahmer: Done, Thanks!

xija/core_2.c
matthewdahmer: Done, thanks!

matthewdahmer:
@taldcroft The second order approximation you use seems to require two separate constants instead of one to generate an equivalent comparison to the fourth order case, so I'm not sure this is entirely a fair comparison, unless we hard code the second order fit approximation constants into the equation (unless I'm missing something). That being said, I think the temperature normalization should solve the instability issue for reasonable temperatures. I did want to test the difference between using different order though, so I built in an option to specify other values with the default being 4.

taldcroft:
@matthewdahmer - I was thinking about hard-coding those "taylor-series" coefficients so there is still just a single overall normalization. I still think it's worth a quick try to see if it works.

matthewdahmer:
@taldcroft I did some testing and both moncar and simplex fitting methods technically work with both 2nd and 4th order cases now, however the simplex method fails to produce a workable model from an accuracy standpoint for the 4th order case. In all cases, the moncar method produces workable and vastly better models than the simplex method, however this is not unique to these models. The moncar method always produces a better model.

matthewdahmer:
Using the moncar fitting method, the 2nd order case produces results that are very close to the 4th order case, with the 4th order case looking slightly better by a very small margin. The following dashboard plots show the results, unfortunately you need a white background to see the numbers due to how JupyterLab generates the plots.

4th Order Case:
image

2nd Order Case:
image

taldcroft:
The results for 2nd order and 4th order look indistinguishable to me. By Occam's razor I'd suggest 2nd order, though it would be nice to see the same fit results for adding the 1st order (linear) coupling plotted in the same way. In any case your idea of allowing an arbitrary polynomial degree seems fine.

matthewdahmer:
Another option could be to include two fit-able parameters instead of one.

- This: C1*T1^N - C2*T2^N
- Instead of this: C1*(T1^N - T2^N)

Such an arrangement would allow the coefficients you used to approximate the fourth order case to be fit to the model, while including the "factor" (original fitted parameter) as a combined value.

taldcroft:
There are many possibilities. I would only choose the more complex one if it provides a noticeably better fit.

BTW, did using a 2nd order in the fitting above mitigate the problems you originally noted in the TWG presentation wrt to fitting / numerical instability? I.e. can you still reproduce the original problems with 4th order, and then do you see the same (or not) with 2nd order?

matthewdahmer:
The temperature normalization seems to have fixed the numerical instability as long as I use the moncar method; this way I can get good fits using either the second or fourth order coupling. Using the second order coupling along with the temperature normalization allows the simplex method to converge on a non-ideal but not unviable model. Using the fourth order coupling (+ temperature normalization) along with the simplex method does not result in a viable model. I'm not seeing the "NaN trap" in any of these cases however, so in a way it seems the temperature normalization seems to have helped in that regard.

matthewdahmer:
Here are all the relevant results in a PDF that's easier to flip through and compare the different results; all these results use normalized temperatures. I added a version that simulates a linear connection (1st order) to see how that compares. The only mathematical difference in this case is the "factor" is the inverse of the tau used in the original built-in coupling used for years.

[radiative_coupling_results.pdf](https://github.com/sot/xija/files/8631438/radiative_coupling_results.pdf)

I'm not seeing a clear benefit to using the `C1*T1^N - C2*T2^N` (2 factor) option, so the original method with normalized temperature and user defined order seems like the best option.

matthewdahmer:
The document above shows both moncar and simplex (simplex for second order case, 4th order simplex didn't produce a viable model - but no NaN trap at least). The levmar method has never worked well, and doesn't even produce a viable model for the current mups2a flight model architecture.

PR #121, opened 2022-05-13
fixed some PEP8 violations and updated 'Hack' to only overwrite final…

jskrist:
… prediction mvals instead of all mvals

## Description
There was a "Hack" in the model.py code to make sure that calculated data was returned for all entries of the `mvals`.  This code essentially copies the second-to-last value (which is computed by the calc function) into the last value as well.  This logic works fine for predicted values, and results in the last two data points being the same exact value.  However, for non-predicted values, this results in overwritten input data.  Most of the time, the input states aren't changing significantly in the last two data point, so this doesn't cause any issues.  However, when the last data point of an input state does change significantly, then the returned `mvals` will no longer match the given inputs.

The fix presented in this PR is to only overwrite the last values for components that are predictions, and to leave non-predicted values alone.

## Interface impacts
This does not change the API, and should have minimal impacts on the user community in most cases.

### Unit tests
I was unable to run the unit test on Windows due to not having the Ska data available locally.
- [X] No unit tests

- [ ] [PLATFORM]:

### Functional tests
This was tested using the FOT MATLAB Tools running MCC and evaluating all the thermal models for a given week.  The thermal predictions were checked against the values without this fix, and the `mvals` from non-predicted components were compared to input states.

In particular, the final value of the `fep_count` component of the ACIS DPA model was compared to the provided inputs from the MAY0222A schedule.

### DIFF
--- a/xija/model.py
+++ b/xija/model.py
@@ -45,9 +45,6 @@ logger = clogging.config_logger('xija', level=clogging.INFO)
 DEFAULT_DT = 328.0
 dt_factors = np.array([1.0, 0.5, 0.25, 0.2, 0.125, 0.1, 0.05, 0.025])
 
-#int calc_model(int n_times, int n_preds, int n_tmals, double dt,
-#               double **mvals, int **tmal_ints, double **tmal_floats)
-
 
 def convert_type_star_star(array, ctype_type):
     f4ptr = ctypes.POINTER(ctype_type)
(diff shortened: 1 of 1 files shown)


xija/model.py
```diff
@@ -692,8 +687,8 @@ def calc(self):
                                      len(self.tmal_ints), dt, mvals,
                                      tmal_ints, tmal_floats)
 
-        # hackish fix to ensure last value is computed
-        self.mvals[:, -1] = self.mvals[:, -2]
+        # hackish fix to ensure last value is a computed value for predicted components
+        self.mvals[:self.n_preds, -1] = self.mvals[:self.n_preds, -2]
```
jskrist, line 691: This is the meat of the PR.  The rest is just PEP8 improvements.

taldcroft: Looks good.

taldcroft (approved):

taldcroft:
Thanks @jskrist !

PR #122, opened 2022-07-15
Fix instabilities in xija_gui_fit, add ability to adjust error limits in histogram plots

jzuhone:
## Description

`xija_gui_fit` occasionally exhibits random crashes, especially when one is attempting to make new plots or pan/zoom within existing plots. This appears to be due to race conditions where an instruction to change a plot is made to a plot object that no longer exists. 

Along with https://github.com/sot/Ska.Matplotlib/pull/25, this PR seeks to avoid these crashes by doing the following:

* Avoid creating and re-creating `Figure` objects as much as possible, instead save the `Figure` object and update its data
* Add a short delay before updating plots after adding limit lines
* Caching rad zone times so they don't need to be created the first time you plot them
* General cleanup and simplification of code

This PR also adds the ability to manually specify the error limits in histogram plots, see the new input boxes in the lower-right of this image:

image

This allows more control over the histogram plots to make them more reproducible. 

### Unit tests

- [x] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

### Functional tests

* Open up `xija_gui_fit`, fit a model, create plots, zoom in and out, pan, re-fit, etc. to see if one gets a crash--no crashes occurred
* Check that changing the error plot limits update plots as desired

### DIFF
--- a/xija/gui_fit/app.py
+++ b/xija/gui_fit/app.py
@@ -89,14 +89,12 @@ class FormattedTelemData:
         return val
 
 
-class FiltersWindow(QtWidgets.QMainWindow):
+class FiltersWindow(QtWidgets.QWidget):
     def __init__(self, model, main_window):
         super(FiltersWindow, self).__init__()
         self.model = model
         self.mw = main_window
         self.setWindowTitle("Filters")
-        wid = QtWidgets.QWidget(self)
-        self.setCentralWidget(wid)
 
         header_font = QtGui.QFont()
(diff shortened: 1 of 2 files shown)


jzuhone:
any comments on this PR?

taldcroft (approved):
The functional testing looks good enough, esp. since this tool is not used directly in mission-critical operations.

taldcroft:
@jzuhone - feel free to merge.

PR #124, opened 2022-09-05
Apply black and isort formatting and modernize config files

taldcroft:
## Description

This does the following:

- Apply [Black](https://black.readthedocs.io/en/stable/) formatting in a few steps
- Apply `isort` to sort imports
- Add `pyproject.toml` to configure `black` and `isort` behavior
- Add `.pre-commit-config.yaml` to allow pre-commit hooks to be installed
- Remove `pytest.ini`. There is a new paradigm where we will use a single global definition of that file, as provided by the `ska_testr` repo. This new `pytest.ini` should be placed in a directory above all your Ska git repos (e.g. `~/git`).
 
## Interface impacts

None

### Unit tests

- [x] Mac

- [ ] [PLATFORM]:

### Functional tests

Confirmed that the pre-commit hooks work as expected.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,7 +1,0 @@
-[pytest]
-filterwarnings =
-    # See https://github.com/numpy/numpy/issues/11788 for why this is benign
-    ignore:numpy.ufunc size changed:RuntimeWarning
-    ignore:the imp module is deprecated in favour of importlib:DeprecationWarning
-    ignore:parse functions are required to provide a named argument:PendingDeprecationWarning
-norecursedirs = .* alt_core build ctypes_example dist doc examples GUI_play ipynb models* nmass* pyside tests_no_pytest tmp twodof www


jzuhone (approved):
totally support this.

matthewdahmer:
@taldcroft I support this. The only hangup I have is the use of double quotes by default; I find the use of single quotes makes code look better, but this isn't really that big of a deal for me. I generally find the code formatting by black to be quite pleasant to read. I didn't see a way in the black api to ignore specific lines that are intentionally formatted in such a way that would otherwise be re-formatted, is this an option?

matthewdahmer (approved):

taldcroft:
OK, I'll re-do this PR after #127 and #125 get merged. I will also put in a pre-commit hook config to make it easier to not forget.

After a little while I got used to the double quotes and now they look better. One thing is that I almost always *type* a single quote and then when I'm done with a section of code I have black fix it with option-shift-f in VS code.

PR #125, opened 2022-09-07
Call a model from another model and use it as a node

jzuhone:
## Description

This PR provides a new class that calls a separate xija model from another model and returns the temperature data as a node, which then can be used as any other component. 

Still need to provide a docstring and some tests, this is just here to start discussion and review.

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

No functional testing.

### DIFF
--- a/xija/component/base.py
+++ b/xija/component/base.py
@@ -338,19 +338,29 @@ class Node(TelemData):
             ax.autoscale()
 
 
-class ModeledData(ModelComponent):
-    def __init__(self, model, name, modeled_msid, model_spec):
+class ModeledData(TelemData):
+    def __init__(self, model, name, modeled_msid, model_spec=None,
+                 set_data_vals=None):
         from xija.model import XijaModel
-        super().__init__(model)
+        from xija.get_model_spec import get_xija_model_spec
+        super().__init__(model, modeled_msid)
         self.modeled_msid = modeled_msid
+        if model_spec is None:
+            model_spec = get_xija_model_spec(name)[0]
         self.model2 = XijaModel(name, start=self.model.datestart,
                                 stop=self.model.datestop,
                                 model_spec=model_spec)
+        if set_data_vals is not None:
+            for comp_name, val in set_data_vals.items():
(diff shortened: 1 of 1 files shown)


taldcroft:
Awesome! One thing that might be good is a `version` parameter to specify the version of the model. I suppose most of the time you want to the latest release (which is the default) but for testing it would be good to be able to pin it.

I've been scratching my head for a few minutes and writing down, then deleting ideas, about how to make this fit in more easily with the existing structure that MATLAB and other Python tools use for initializing models. Basically there is no precedent for nested/embedded components, so I'm a bit worried about this requiring substantial upstream changes to use.

Maybe this model itself is OK and all that needs work is the HRC model to use this new model in a way that is consistent with existing APIs. So perhaps this PR needs a companion demonstration HRC model that "just works" in xija_gui_fit. At first glance that would mean that ACIS components like `ccd_count` or `clocking` would be treated as formal parameters of the HRC model but would get passed through to the embedded DPA model. Maybe the HRC model `comp` could be a property that does this delgation for the right components? Just thinking out loud.

I'm also curious if this serializes to JSON and Python properly.

jzuhone:
@taldcroft Here's a working example in `xija_gui_fit` for the HRC model Dan passed to me:

image

Note the presence of the 1DPAMZT model data. Histogram:

image

Here is the JSON file you need for this:

https://gist.github.com/jzuhone/f34f60a1f3f6052c5a4ce900a6148b7c

taldcroft:
@jzuhone - unfortunately I'm no longer able to run xija_gui_fit, see #126.

jzuhone:
@taldcroft I have a fix for this incoming.

taldcroft:
Hum, just noticed it works fine for an old DPA model, so maybe it is particular to this spec file?

jzuhone:
I rebased this on top of the merged #127

jzuhone:
Closing this because we decided not to use it.

PR #128, opened 2022-09-23
create a generic solar heating class which is SIM-Z dependent

jzuhone:
## Description

The `AcisPsmcSolarHeat` class provides separate sets of pitch bins for various ranges of SIM-Z position. 
This PR refactors that class into a new class, `SimZDepSolarHeat`, and then subclasses `AcisPsmcSolarHeat` and `HrcCeaSolarHeat`, which supply their own SIM-Z ranges. 

## Interface impacts

The `AcisPsmcSolarHeat` class should be completely unaffected, but will test. 

### Unit tests

- [ ] No unit tests
- [ ] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

Will test outputs of existing PSMC model against this, should be identical.

### DIFF
--- a/xija/component/heat.py
+++ b/xija/component/heat.py
@@ -768,13 +768,13 @@ class DetectorHousingHeater(TelemData):
     def __str__(self):
         return 'dh_heater'
 
-
-class AcisPsmcSolarHeat(PrecomputedHeatPower):
-    """Solar heating of PSMC box.  This is dependent on SIM-Z"""
-    def __init__(self, model, node, pitch_comp='pitch', simz_comp='sim_z',
-                 dh_heater_comp='dh_heater', P_pitches=None,
-                 P_vals=None, dPs=None, var_func='linear',
-                 tau=1732.0, ampl=0.05, epoch='2013:001:12:00:00', dh_heater=0.05):
+class SimZDepSolarHeat(PrecomputedHeatPower):
+    """SIM-Z dependent solar heating"""
+    def __init__(self, model, node, simz_lims, instr_names, 
+                 pitch_comp='pitch', simz_comp='sim_z', 
+                 dh_heater_comp='dh_heater', P_pitches=None, P_vals=None, 
+                 dPs=None, var_func='linear', tau=1732.0, ampl=0.05, 
+                 epoch='2013:001:12:00:00', dh_heater=0.05):
         ModelComponent.__init__(self, model)
         self.n_mvals = 1
         self.node = self.model.get_comp(node)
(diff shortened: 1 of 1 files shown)


jzuhone:
@taldcroft @matthewdahmer @drxre check this out, still need to test it

xija/component/heat.py
```diff
@@ -869,18 +866,60 @@ def plot_solar_heat__pitch(self, fig, ax):
...
+    """Solar heating of PSMC box.  This is dependent on SIM-Z"""
+    def __init__(self, model, node, pitch_comp='pitch', simz_comp='sim_z',
+                 dh_heater_comp='dh_heater', P_pitches=None,
+                 P_vals=None, dPs=None, var_func='linear',
+                 tau=1732.0, ampl=0.05, epoch='2013:001:12:00:00', dh_heater=0.05):
+        simz_lims = [[-400000.0, -85000.0],  # HRC-S
+                     [-85000.0, 0.0],        # HRC-I
+                     [0.0, 83000.0],         # ACIS-S
+                     [83000.0, 400000.0]]    # ACIS-I
+        instr_names = ['hrcs', 'hrci', 'aciss', 'acisi']
```
taldcroft, line 892: What about making these class attributes (and maybe immutable tuples as well). Then the subclasses have only changes to the class attributes and the `__str__` but no `__init__`. The base class `SimZDepSolarHeat` would have those attributes set to `None` which means that this is not directly usable as a model component.

Otherwise this looks great at first glance, definitely a good idea.

xija/component/heat.py
```diff
@@ -869,18 +866,60 @@ def plot_solar_heat__pitch(self, fig, ax):
...
+                         pitch_comp=pitch_comp, simz_comp=simz_comp, 
+                         dh_heater_comp=dh_heater_comp, P_pitches=P_pitches,
+                         P_vals=P_vals, dPs=dPs, var_func=var_func, tau=tau,
+                         ampl=ampl, epoch=epoch, dh_heater=dh_heater)
+
     def __str__(self):
         return 'psmc_solarheat__{0}'.format(self.node)
 
 
+class HrcCeaSolarHeat(SimZDepSolarHeat):
```
taldcroft, line 903: In the spirit of component re-usability, maybe this should be given a more generic name? `SimZDepHrcISAcisSolarHeat` or whatever.

xija/component/heat.py
jzuhone: I agree

jzuhone:
Tests pass

xija/component/heat.py
```diff
@@ -869,7 +874,8 @@ def plot_solar_heat__pitch(self, fig, ax):
                 lines[i].set_data(self.P_pitches, P_vals[instr_name])
                 # lines[i * 2 + 1].set_data(self.P_pitches, P_vals[instr_name], '-b')
         else:
-            for instr_name, color in zip(self.instr_names, colors):
+            for i, instr_name in enumerate(self.instr_names):
+                color = colors[i]
```
taldcroft, line 878: Why this change?

taldcroft (commented):
Looks good with just one question. The only other thing this needs is a functional or unit test of having different numbers of P and dP pitch bins. I think the existing testing doesn't cover that case.

xija/component/heat.py
jzuhone: Done

xija/component/heat.py
jzuhone: Originally the number of colors was equal to the number of SIM-Z position ranges, which was 4. This allows it to be any number >= 2, but if we wanted it to be larger than 4 we'd need to make the `color` list longer.

jzuhone:
@taldcroft I agree re: testing--I have a working model for the ACIS FP using one of the new classes with fewer dP than P--if you want I can show the results somehow.

jzuhone:
This is ready to go.

xija/component/heat.py
taldcroft: `zip` will stop when the smallest list finishes, so the original was fine. To be really correct here you could do:
```
zip(self.instr_names, itertools.cycle(colors))
```

taldcroft (commented):
Post-merge review.

For the record can you state the testing for [bddaa56](https://github.com/sot/xija/pull/128/commits/bddaa56b272c88c9d9c5cee70ecf3b21d645ca29)? I saw something on slack but not sure.

taldcroft:
@jzuhone - FYI my 👍 was just acknowledging that this is ready for review so the merge was a little premature, but OK. I just changed the master branch protection rules to require a review for merge. 😺

PR #123, opened 2022-09-05
Improve performance of bad_times and mask_times

taldcroft:
## Description

This fixes a performance issue note on Slack in #thermal_working_group:

*I am doing some work that requires me to run the thermal models many, many times in a row, and was profiling my code when I noticed that the pm2thv1t model is running significantly slower than the pm1thv2t model, e.g. in my MATLAB setup:*
- pm1thv2t takes ~032ms to run, whereas
- pm2thv1t takes ~643ms to run

This PR vectorizes the bad times processing and minimizes the number of date --> time (CXC second) conversions. It also factors this processing out into a separate function (working to improve separation in this code).

Some points of note:
- The new code assumes that the `bad_times` in a JSON model spec file are in standard Year-day-of-year date format. There was another part of the code already making this assumption so I think this is OK but this needs review.
- This code changes the behavior of the `mask_*` attributes. Previously these values were set corresponding to every `bad_times` interval in the spec file, including those that are outside of the model date range. While this was not a problem per se it was likely impacting performance. See the functional testing section for more.

## Interface impacts

None.

### Unit tests

- [x] Mac

Independent check of unit tests by [REVIEWER NAME]
- [ ] [PLATFORM]:

### Functional tests

I ran this script on master (version 4.26.1) and this branch (4.27.1.dev1+g19aed41).
```python
from time import time

import xija
from xija.get_model_spec import get_xija_model_spec

print(xija.__version__)


def print_mdl_times(mdl):
    print(f"{len(mdl.bad_times)=}")
    print(f"{len(mdl.bad_times_indices)=}")
    print(f"{len(mdl.mask_times)=}")
    print(f"{len(mdl.mask_time_secs)=}")
    print(f"{len(mdl.mask_times_bad)=}")


def timer_func(func):
    # This function shows the execution time of
    # the function object passed
    def wrap_func(*args, **kwargs):
        t1 = time()
        result = func(*args, **kwargs)
        t2 = time()
        name = kwargs["model_spec"]["name"]
        datestart, datestop = args[1:3]
        print(f"{name=} {datestart=} {datestop=} executed in {(t2-t1) * 1000:.1f} ms")
        print_mdl_times(result)
        return result

    return wrap_func


# No bad times in spec1
spec1, version = get_xija_model_spec("pm1thv2t", version="3.40.2")

# 5876 bad times in spec2
spec2, version = get_xija_model_spec("pm2thv1t", version="3.40.2")

XijaModel = timer_func(xija.XijaModel)

[...]

PR #130, opened 2022-10-04
Fix problem of dropping bad_times in spec file

taldcroft:
## Description

This fixes the issue noted on slack that after #123, any bad times in a model spec file that were not covered by the fitting range in `xija_gui_fit` would be dropped from the spec file.

## Interface impacts

None.

### Unit tests

- [ ] No unit tests
- [x] Mac
- [ ] Linux
- [ ] Windows

- [ ] [PLATFORM]:

### Functional tests

I made a copy of the flight `aca_spec.json` file and added a new bad time interval (not strictly necessary but I did it):
```
        [
            "2020:165:14:10:00",
            "2020:170:00:00:00"
        ]
```
Then using both `master` branch and this branch, I did:
```
python -m xija.gui_fit.app aca_spec.json --stop 2020:190 --days 60
```
In both cases:
- Confirm seeing the two expected bad time intervals marked in light blue on the plot.
- Confirm that masked values are hidden in the residual plot.
- Freeze all parameters and thaw a couple of solar heat P parameters
- Fit and confirm things look OK with no obvious influence of the bad time intervals.
- Save spec file as JSON

Using the `master` branch I saw that all the saved bad times outside of the fit range were missing.
Using this branch I saw that the saved bad times matched exactly the original bad times.

### DIFF
--- a/xija/gui_fit/plots.py
+++ b/xija/gui_fit/plots.py
@@ -49,7 +49,7 @@ def calcquantiles(errors):
     error :
         model errors (telemetry - model)
     errors :
-        
+
 
     Returns
     -------
@@ -76,7 +76,7 @@ def calcquantstats(Ttelem, error):
     -------
     type
         coordinates for error quantile line
-        
+
         This is used for the telemetry vs. error plot (axis 3).
 
     """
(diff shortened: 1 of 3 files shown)


matthewdahmer (approved):
Tested via the MUPS2A model (`pm2thv1t_spec_matlab.json`) by performing a model fit for only the last 100 days, writing out the resulting model spec to a file, and ensuring the bad times for dates prior to the last 100 days persist.

jzuhone (approved):

PR #134, opened 2023-08-12
Refactor get_model_spec to use ska_helpers.chandra_models

taldcroft:
## Description

This refactors the `get_model_spec` module to use `ska_helpers.chandra_models`.

## Interface impacts

None, this is expected to be API compatible with the previous version.

### Unit tests

- [x] Mac

Independent check of unit tests by Jean
- [x] Linux

### Functional tests

No functional testing.

### DIFF
--- a/xija/get_model_spec.py
+++ b/xija/get_model_spec.py
@@ -5,17 +5,14 @@ Get Chandra model specifications
 import contextlib
 import json
 import os
-import platform
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


jeanconn:
I note that removal of the repo-tip check (already removed in ska_helpers) could count as an interface / api change.

jeanconn (approved):

PR #146, opened 2025-09-03
xija_gui_fit enhancements

jzuhone:
## Description

This PR adds a number of enhancements to `xija_gui_fit`, cleans up the code a little, and adds documentation. The extensive list of new features is:

* Added a legend and autoscaling for the solarheat pitch plots
* Added a plot for the ACIS power state coefficients
* Added a native menu bar for the application, and moved several of the buttons on the main window to it
* Moved the "fit status" to the status bar at the bottom of the main window
* Writing the model + data time series output to an ASCII file now uses the ECSV format
* Added functionality to make a plot showing the evolution of the fit statistic with number of iterations
* Made the box containing the main window plots scrollable
* Made changes to several dialogs which took input start and stop times, ensuring that these were valid and raising an informative error if they weren't
* Added a title for the column with the sliders in the parameter table

I have added complete documentation for `xija_gui_fit` in this PR as well, which also is the best place to take a look at the new features. I have a rendered version of the docs up at https://cxc.cfa.harvard.edu/acis/xija/gui_fit.html.

Otherwise, I made some stylistic / cleanup / bugfix changes:

* Removed subclassing from `object`
* The `XijaModel` was being added as an attribute to a number of subclasses when it wasn't necessary, tried to pare this down
* Made a function name more pythonic
* Modernized calls to `super().__init__()`
* Removed some dead code
* Added PyCharm's `.idea` directory to `.gitignore`
* Fixed a `help` string in the `ArgumentParser`
* Updated code for more recent versions of Matplotlib (post 3.7.0)
* Removed unused arguments and out-of-date comments
* Made it so that the old ACIS FP limits did not appear on various plots

## Interface impacts
Technically speaking, given that `xija_gui_fit` is a "graphical user interface", there are lots of interface changes, but they are all documented here, and since it only fits thermal models, nothing here affects flight operations immediately or directly. 

## Testing

Unit tests are not really feasible for `xija_gui_fit`, so we must rely on extensive functional testing. 

### Unit tests

- [x] No unit tests

### Functional tests

Carried out a full fitting workflow on both macOS and Linux, exercising all options.


jeanconn:
We've updated the ska3 template ruff-base to at least ignore PLC0415.

